PR #8, opened 2016-09-09
Only load pingbacks when needed

### LINKED ISSUE #6: Only load pingback URL when needed
Only load the pingback URLs when it's a singular page and pingbacks are open.

HardeepAsrani:
This condition will only load pingback URL when it's a singular page template and the pingbacks are accepted. Close #6

### DIFF
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -95,7 +95,7 @@ function twentyseventeen_is_frontpage() {
  */
 function twentyseventeen_pingback_header() {
 	if ( is_singular() && pings_open() ) {
-		echo '<link rel="pingback" href="', bloginfo( 'pingback_url' ), '">' . "\n";
+		printf( '<link rel="pingback" href="%s">' . "\n", get_bloginfo( 'pingback_url' ) );
 	}
 }
 add_action( 'wp_head', 'twentyseventeen_pingback_header' );
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -89,3 +89,13 @@ function twentyseventeen_is_frontpage() {
 		return false;
 	endif;
 }
+
+/**
+ * Add a pingback url auto-discovery header for singularly identifiable articles.
+ */
+function twentyseventeen_pingback_header() {
+	if ( is_singular() && pings_open() ) {
+		echo '<link rel="pingback" href="', bloginfo( 'pingback_url' ), '">' . "\n";
+	}
+}
+add_action( 'wp_head', 'twentyseventeen_pingback_header' );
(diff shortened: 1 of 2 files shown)


HardeepAsrani:
Other solution could be to load it outside the header.php, like _s does: https://github.com/Automattic/_s/blob/ff4b374d6bf289eb602e9cebba369077dd446050/inc/extras.php#L34

But that doesn't seem right to me. The current approach seems way better.

swissspidy:
The way _s does it makes it easier for users to remove it in a child theme, without needing to override the whole `header.php` file. Removing the `<link>` tag would be as easy as doing `remove_action( 'wp_head', 'twentyseventeen_pingback_header' );`

HardeepAsrani:
@swissspidy So should I send that instead?

HardeepAsrani:
@swissspidy I do have a question, why would a user want to remove that tag? The way I sent a pull, if the pingbacks are disabled then it will be removed automatically.

trsenna:
I think this peace of code is better within the functions (like _s), just to keep the header.php as cleaner as possible.

swissspidy:
@HardeepAsrani Because it's useless in theory when the `X-Pingback` HTTP header is sent, someone wants to disable pingbacks completely, someone wants to save a few bytes of the HTTP request, … lots of scenarios.

HardeepAsrani:
@swissspidy Thanks for the info. I'll update the commit. :)

HardeepAsrani:
@swissspidy Hope it's fine. :)

inc/extras.php
```diff
@@ -89,3 +89,13 @@ function twentyseventeen_is_frontpage() {
 		return false;
 	endif;
 }
+
+/**
+ * Add a pingback url auto-discovery header for singularly identifiable articles.
+ */
+function twentyseventeen_pingback_header() {
+	if ( is_singular() && pings_open() ) {
+		echo '<link rel="pingback" href="', bloginfo( 'pingback_url' ), '">' . "\n";
```
davidakennedy: @HardeepAsrani Why the `"\n"` here? Seems not needed.

swissspidy: Well, with the `\n` it properly adds a newline so the HTML output looks cleaner.

I rather wonder about the commas there.

davidakennedy: Yeah, good call about the commas. I think this would be better:

`printf( '<link rel="pingback" href="%s">' . "\n", get_bloginfo( 'pingback_url' ) );`

That, the concatenation is more natural. And `bloginfo` outputs, it doesn’t return so this wouldn't echoing on an echo.

@HardeepAsrani Want to update the PR?

HardeepAsrani: @davidakennedy I'll update. :)

davidakennedy:
Thanks @HardeepAsrani!

HardeepAsrani:
Glad to be of help. @davidakennedy :)

PR #17, opened 2016-09-10
Use absint to validate page IDs in customizer

claudiosanches:
`intval()` already test if "is numeric" and return `0` if needed.

```
$ wp shell
wp> intval( 'WordPress' )
=> int(0)
```

But since it validate post/page IDs, it should make sure that the numbers are positives too.
So we can simple use the [`absint()`](https://developer.wordpress.org/reference/functions/absint/) function.

wp.org username: claudiosanches

### DIFF
--- a/inc/customizer.php
+++ b/inc/customizer.php
@@ -64,7 +64,7 @@ function twentyseventeen_customize_register( $wp_customize ) {
 
 	$wp_customize->add_setting( 'twentyseventeen_panel1', array(
 		'default'           => false,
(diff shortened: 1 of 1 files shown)


davidakennedy:
Hey @claudiosmweb! Thanks for the pull request! Nice catch, I'd like to get this in – can you refresh it?

claudiosanches:
@davidakennedy updated, thanks :)

davidakennedy:
Thanks @claudiosmweb!

PR #37, opened 2016-09-10
Full width page

WPDevHQ:
Adding an actual full width page template.


melchoyce:
Thanks for your suggestion! At this time we're not planning to include a "full-width" page template, since pages won't be including sidebars. You can see the regular page template here: https://cloudup.com/iYKCaBTfZMt

lukecav:
Okay makes sense thanks.

WPDevHQ:
I don't think the aproach was understood correctly...

For design purposes and for many scenarios a full width does not mearely mean not having a sidebar on the page. Its about sitting the content centered on the page with minimal white space!

The current design does not allow for this.

melchoyce:
Hi @WPDevHQ, if you go into the Customizer and go to theme options, there's an option for a one-column layout.

The layout isn't specifically "full-width" because that would negatively impact readability. Allowing text to span the full width of your screen makes it significantly harder for people to read your content. Line lengths should sit around [50-75 characters per line](http://baymard.com/blog/line-length-readability) for optimal readability.

daveit:
>> The layout isn't specifically "full-width" because that would negatively impact readability. 

Content does not have to be one column across the entire width - Many of the page builders (and multi column plugins) provide the ability to have multiple columns with varying content in them...

PR #47, opened 2016-09-10
Add end comment

bappi:
add an end comment in line 7.. end of class `author-avatar`


grappler:
@B-07 Could you edit the title to mention the file? In the ideal world you would create one pull request for all of the files together as it fixes one issue.

bappi:
ture. to be honest I didn't think I'll find as much. the last one contains multiple files

PR #58, opened 2016-09-10
Fix issue with social menu using default left-margin

pattonwebz:
It thows off the center alignment of the social menu. 

WordPress.org username - williampatton


laurelfulford:
Thanks!

PR #61, opened 2016-09-11
Remove undefined variable $post

### LINKED ISSUE #56: Undefined variable: post
These notices are not visible as the are hidden behind the header image and only noticed them after removing the image.

Notice: Undefined variable: post in F:\Users\Zulf\Desktop\UniServerZ\www\seventeen\wp-content\themes\twentyseventeen-master\inc\extras.php on line 39

Notice: Trying to get property of non-object in F:\Users\Zulf\Desktop\UniServerZ\www\seventeen\wp-content\themes\twentyseventeen-master\inc\extras.php on line 39

Edit: Adding `global $post;` fixes this but I'm not sure if that is the ideal solution.

grappler:
- Make has_header_image() check stricter

fixes #56

w.org username: grapplerulrich

### DIFF
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -35,8 +35,7 @@ function twentyseventeen_body_classes( $classes ) {
 	}
 
 	// Add class if no custom header or featured images
-	$header_image = get_header_image();
-	if ( '' == $header_image && ( ! has_post_thumbnail( $post->ID ) || is_home() ) ) {
+	if ( ! has_header_image() && ( ! has_post_thumbnail() || is_home() ) ) {
 		$classes[] = 'no-header-image';
 	}
 


inc/extras.php
```diff
@@ -35,8 +35,7 @@ function twentyseventeen_body_classes( $classes ) {
 	}
 
 	// Add class if no custom header or featured images
-	$header_image = get_header_image();
-	if ( '' == $header_image && ( ! has_post_thumbnail( $post->ID ) || is_home() ) ) {
+	if ( false === has_header_image() && ( ! has_post_thumbnail() || is_home() ) ) {
```
ahmadawais: @grappler This could be done like this 

``` php
if ( ! has_header_image() && ( ! has_post_thumbnail() || is_home() ) ) {
```

just to be consistent with the next check what do you think?

ahmadawais:
@grappler This could be done like this 

```php
if ( ! has_header_image() && ( ! has_post_thumbnail() || is_home() ) ) {
```
just to be consistent with the next check what do you think? 

W.org Username: `mrahmadawais`

inc/extras.php
grappler: It would be good to be consistent. `false ===` makes the check stricker.

ahmadawais: I think you should use one of them for every check in the statement

``` php
if ( false === has_header_image() && ( false === has_post_thumbnail()
```

OR

``` php
if ( ! has_header_image() && ( ! has_post_thumbnail() 
```

Guess, now it would be more clear what I was talking about.

ahmadawais: That's why I said, simpler is better.

swissspidy: Strict checks are only really useful/necessary when a function can return different types, e.g. a bool or a string, depending on context. Since `has_header_image()` only returns a bool, a simpler check should be preferred.

grappler: Thanks guys I have pushed the changes. Should be good now.

ahmadawais: 👍

ahmadawais (approved):

PR #74, opened 2016-09-20
Replace front page panels with a widget area.

bradyvercher:
Here's an alternative concept to the menu-based approach in #62 for discussion.

A lot of themes already use widget areas for front page content, but it's always been difficult to account for the number of possible widgets from a design perspective. This approach adds an `allowed_widgets` argument to `register_sidebars()` to whitelist the widgets that can be added to a widget area.

The section is moved from the Widgets panel to the Theme Options panel in the Customizer where the front page panels are currently located. There are a few tweaks that could be made in core to further differentiate front page content from a normal widget area, like allowing the "Add a Widget" button text to be updated. The Widgets screen would also need to account for the whitelist somehow, but this should serve as a quick concept.

The benefit is that it requires very little code and is already a familiar concept. There are some limitations compared to #62, but I don't think that's necessarily a bad thing at this stage.


inc/customizer.php
```diff
@@ -15,6 +15,9 @@ function twentyseventeen_customize_register( $wp_customize ) {
 	$wp_customize->get_setting( 'blogdescription' )->transport  = 'postMessage';
 	$wp_customize->get_setting( 'header_textcolor' )->transport = 'postMessage';
 
+	// Move the Front Page widget area to the Theme Options panel.
+	$wp_customize->get_section( 'sidebar-widgets-front-page' )->panel = 'twentyseventeen_options_panel';
```
grappler: This is creating an error `Warning: Creating default object from empty value in/twentyseventeen/inc/customizer.php on line 19`

bradyvercher: Huh, I guess the section doesn't always exist when the controls are registered. I didn't spend much time on the code since this is more about exploring the concept, but this should be fixed in cc2b7d6. Thanks for letting me know about that!

celloexpressions:
As with #62, the section should remain under widgets. There also isn't a need for it to be nested under another panel if it were brought out. API-wise, we can't reliably account for widgets not living in the panel currently, and there is likely a fair amount of JS that assumes widget areas are in the widgets panel (even more so for menus). We could try to loosen some of those restrictions in core and add an API for changing the button label (not to mention all of the labels in the shared add-widget section), but this would very likely lead to backwards compatibility nightmares in the future for core.

bradyvercher:
Moving the widget area was done for a few reasons:
1. To demonstrate how it'd work in contrast to the existing panels solution
2. Alleviate the discoverability issues raised in [#37974](https://core.trac.wordpress.org/ticket/37974)
3. And most importantly, to differentiate it from a regular widget area in order to reset user expectations. We used this concept before widgets were added to the Customizer and liked it from a control perspective, but users are accustomed to dropping any available widget in a widget area, which is understandable.

Ideally, there shouldn't be anything in core that's absolutely tied to the Widget panel markup. From the bit of testing I've done, I haven't noticed any issues from moving the section, but that doesn't mean there aren't any.

Whether the section exists at the top level or in a separate panel doesn't really matter much to me, but I think there's promise in the concept and differentiating it from a regular widget area is important.

celloexpressions:
My concerns above are more for technical reasons. However, I also want to point out that for these sorts of issues we should look at how to expand widgets and menus to be more flexible and better communicate how they're used rather than trying to mimic their functionality but pull out other things and present them as separate features. One obvious way to improve this would be to create ways in core to define what types of menu items or widgets can be added to a given area/location.

joyously:
As someone mentioned on a different issue, maybe what we are really looking at is a front page template, and the options for that template are in a meta box under the page editor. Sounds good to me.... like these types of options (content-related) don't belong in the Customizer, which is about Appearance and not Content.
I still don't understand why the push to rewrite all of WordPress under the Customizer.

davidakennedy:
Lots of awesome work and ideas have been done here. Thank you, everyone!

In today's weekly meeting for Twenty Seventeen, the group decided to make [#37974](https://core.trac.wordpress.org/ticket/37974) the master ticket for making it easier to create a custom home page full of content from different sources.

See: https://make.wordpress.org/core/2016/09/23/twenty-seventeen-meeting-notes-sept-23-2016/

So let's move all discussion around there. That way, more people see it and we can focus and find the best solution.

Issue #12, opened 2016-09-09
CSS review


laurelfulford:
The theme's CSS needs to be reviewed to make sure it follows best practices. 

This should be done once #4 is complete.

Presskopp:
I wonder if these lines from L.2449 like
_.main-navigation ul ul ul a:hover,_
can be shortened using odd/even and the like (selectors). Or is this needed because of BC?
Sorry if I'm spamming.

marcustisater:
I'm up for it! So just to make sure, are we following the [WordPress CSS Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/)?

grappler:
Yes, we are following the coding standards.

marcustisater:
Thanks man, I will have a look and leave PR.

juanfra:
@marcustisater check this out: https://rezzz.com/php-codesniffer-with-wordpress-coding-standards-and-atom/ - It's a great tool to check that.

celloexpressions:
I saw a number of minor issues when I went through the whole stylesheet for colors yesterday. There were a couple of duplicate selector blocks in there I think.

marcustisater:
Thanks for the heads up, expect a PR in a couple of days 😄

joyously:
One best practice is to keep the theme's selectors at a consistent specificity, so that the cascade makes sense and overrides are easier. Most selectors in this theme use classes, but a few are using IDs, which are more specific. 
`#page`
`#masthead .wrap`
`body:not(.twentyseventeen-front-page) #comments`
`.single-post:not(.has-sidebar) #primary, .page.page-one-column #primary,
.home.blog:not(.has-sidebar) #primary`
`#comments`
`#secondary`
`.twentyseventeen-front-page.twentyseventeen-customizer #primary article.panel-placeholder`
`.has-sidebar:not(.error404) #primary`
`.has-sidebar #secondary`
`.error404 #primary`
`body.has-sidebar.error404 #primary .page-header,
    body.page-two-column #primary .entry-header`
`.blog:not(.has-sidebar):not(.home) #primary article,
    .archive:not(.has-sidebar) #primary article,
    .search:not(.has-sidebar) #primary article,
    .error404:not(.has-sidebar) #primary .page-content,
    .error404.has-sidebar #primary .page-content,
    body.page-two-column #primary .entry-content`
`.has-sidebar #primary .entry-content blockquote.alignright.below-entry-meta`
[...]

PR #76, opened 2016-09-20
Update style.css

SGr33n:
The header title line height (site title) is different if you are on the home page or on another page. This is due to the fact that on the home page we have the title inside a H1. With this addition the header title will always have the same line height.


josephfusco:
Cross referencing #12 so this appear under it

davidakennedy:
Nice catch @SGr33n – thank you for your contribution!

PR #81, opened 2016-09-21
Escape classes in body_class filter

jigneshnakrani088:
Escape classes added through **twentyseventeen_body_classes** by _**esc_attr**_ function

### DIFF
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -10,7 +10,7 @@
 /**
  * Adds custom classes to the array of body classes.
  *
- * @param array $classes Classes for the body element.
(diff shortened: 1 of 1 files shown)


grappler:
Why do we need to escape the classes? The classes that we are adding are hardcoded and not coming from an unknown source.

If we were to escape the body classes then it should be done as late as a possible before output in the `header.php`.

ahmadawais (commented):
Squash the commits I think. Also, I agree with Ulrich, why is there a need for escaping hard coded data.

Issue #68, opened 2016-09-17
Several Minor Documentation Issues


ahmadawais:
Throughout the theme, there are several minor documentation issues as per the [WP PHP Documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#6-file-headers). I plan to fix them with a PR.
- [x] FIX: There should be no period for file headers' summary 
- [x] FIX: More than two times a parameter has not been documented
- [x] FIX: Many files have no file DocBlock
- [x] NEW: @pacakge WordPress
- [x] NEW: @subpackage Twenty_Seventeen ( I'd opt for a new package name which should be a single word instead of two words.
- [x] NEW: @since (I'd opt for adding @since 1.0.0 to each and every file head)

Looking forward!

davidakennedy:
Thanks for the issue and pull request @ahmadawais! This has been on my list, so happy to have help with it. I'll leave some specific feedback on the PR itself.

ahmadawais:
@davidakennedy Great :)

I left some questions about the format for each file header in the PR. Do take a look. 

I'd like to contribute most of my time here with TwentySeventeen by keeping a check on each and every file to help it conform with the WP Docs/Coding standards. This will help us set the right expectations and example for new developers as well.

PR #69, opened 2016-09-17
FIX: No period for file headers #68

ahmadawais:
There should be no period for file headers' summary as per the [WP PHP Documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#6-file-headers).

Fixes: #68

W.org Username: `mrahmadawais`

PR #82, opened 2016-09-22
FIX: No period for file headers 💯

ahmadawais:
There should be no period for file headers' summary as per the [WP PHP Documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#6-file-headers).

Fixes: #68 and #69 (duplicate)

W.org Username: `mrahmadawais`


ahmadawais:
@davidakennedy Would love to get your feedback on this.

Issue #72, opened 2016-09-19
 Add version number for all single PHP files for better handling child themes


Zodiac1978:
I've made a plugin called child theme check to keep track of changes in the parent theme (See: https://wordpress.org/plugins/child-theme-check/)

This would prevent security holes / missing features through unchanged files in child themes after an (security/feature) update of the parent theme.

There are slides from my talk at WordCamp Cologne which are explaining the issue that is solved by the plugin:
http://de.slideshare.net/zodiac1978/the-child-theme-dilemma-en

WooCommerce has a similar system to show outdated plugin template files:
http://speakinginbytes.com/2014/02/woocommerce-2-1-outdated-templates/

There is Pull Request for the Theme Check plugin, too:
https://github.com/Otto42/theme-check/issues/115

I really would like to see this in Twenty Seventeen. :)

Already discussed here: https://core.trac.wordpress.org/ticket/36972

2ndkauboy:
+1

Drivingralle:
+1

dimbak:
+1

christopherkurth:
+1

davidakennedy:
@Zodiac1978 Thanks for the issue and pull request!

Just so I understand correctly, would these version numbers be updated only when a file is updated? Or updated every release?

Zodiac1978:
@davidakennedy Thanks for asking! These version numbers would be updated only when the file is updated. With this information we can check if a file is outdated in a child theme.

arkonisus:
+1

davidakennedy:
@Zodiac1978 Cool, that makes sense. I'd like to merge the PR. Can you update it.

Just drop the new header below the last one in each file: https://github.com/WordPress/twentyseventeen/blob/master/404.php#L9

Zodiac1978:
@davidakennedy Great to hear that. I've updated the PR. :)

davidakennedy:
Left a comment on the PR. 😄

PR #86, opened 2016-09-22
Add @version to php doc block

Zodiac1978:
See Issue #72


Zodiac1978:
Updated. See #72

davidakennedy:
@Zodiac1978 Thanks for the update! One bit of feedback, and a question:
- Can you change `1.0.0` to `1.0`? Default themes in general don't do the minor release number, so that third number isn't needed. Sorry, should have mentioned that.
- Question: It looks like you've stuck to just the main template files? Should we do this will all files? Like the partials included in `components`?

Thanks!

Zodiac1978:
@davidakennedy I've changed the version number according to your feedback. I missed the partials totally. They are included with `get_template_part` so they can be changed via child theme, so I would want to add versions for them too.

The files in `inc` are included via `get_template_directory` so they are not meant to be overridden via child theme and the `functions.php` is not overridden but an addition - so for my intended use case it is not necessary to add version number for these files.

davidakennedy:
@Zodiac1978 Right, that makes sense. I really only meant the partials that had template code. 😄  Thanks for the update!

bueltge:
Small hint to the version format, like 1.0.0 vs 1.0. I think all public projects should use the semantic versioning, not a custom style. Is much easier to document, understand and also important, easier to contributing. More about semantic versioninig is [on the page](http://semver.org/) to this topic.

PR #88, opened 2016-09-22
Remove else statement after return. Fixes #87

### LINKED ISSUE #87: Redundant else statement after return
Consider removing the redundant else statement after `return`

example:

```
if ( is_numeric( $input ) ) {
    return intval( $input );
} else {
    return 0;
}
```

_should_ read

```
if ( is_numeric( $input ) ) {
    return intval( $input );
}

return 0;
```

ghost:
WordPress.org username: implenton

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -124,6 +124,7 @@ function twentyseventeen_edit_link( $id ) {
 		'<span class="edit-link">',
 		'</span>'
 	);
+
 	return $link;
 }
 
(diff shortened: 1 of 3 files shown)


inc/template-tags.php
```diff
@@ -124,6 +124,7 @@ function twentyseventeen_edit_link( $id ) {
 		'<span class="edit-link">',
 		'</span>'
 	);
+
```
swissspidy, line 127: Unrelated change, but OK I guess.

swissspidy (approved):
Looks good. This more closely follows what's being done in other default themes and core itself, so 👍

PR #89, opened 2016-09-23
Rolling out Twenty Seventeen design

laurelfulford:
The theme started as a fork of Lodestar. Changes will be made over multiple commits to update the design to match the Twenty Seventeen mockups.

See #4.


comments.php
```diff
@@ -28,25 +28,13 @@
 		<h2 class="comments-title">
 			<?php
 				printf( // WPCS: XSS OK.
-					_nx( 'One thought on &ldquo;%2$s&rdquo;', '%1$s thoughts on &ldquo;%2$s&rdquo;', get_comments_number(), 'comments title', 'twentyseventeen' ),
+					_nx( 'One Reply to &ldquo;%2$s&rdquo;', '%1$s Replies to &ldquo;%2$s&rdquo;', get_comments_number(), 'comments title', 'twentyseventeen' ),
```
grappler, line 34: Is there a reason why the R is capital in reply?

components/page/content-front-page-panels.php
```diff
@@ -47,12 +44,38 @@
...
 
+			<?php
+			// Show recent blog posts if is blog posts page (Note that get_option returns a string, so we're casting the result as an int)
+			if ( get_the_ID() === (int) get_option( 'page_for_posts' )  ) : ?>
+
+				<?php // Show four most recent posts
+				$recent_posts = new WP_Query( array(
+					'posts_per_page'      => 3,
+					'post_status'         => 'publish',
+					'ignore_sticky_posts' => true
```
grappler: As there is no pagination we should be able to add `'no_found_rows' => true` which will improve the query performance. https://10up.github.io/Engineering-Best-Practices/php/#performance

functions.php
```diff
@@ -225,7 +220,12 @@ function twentyseventeen_scripts() {
 
 	wp_enqueue_script( 'twentyseventeen-navigation', get_template_directory_uri() . '/assets/js/navigation.js', array(), '20151215', true );
 
-	wp_enqueue_script( 'twentyseventeen-global', get_template_directory_uri() . '/assets/js/global.js', array( 'jquery' ), '20151215', true );
+	wp_localize_script( 'twentyseventeen-navigation', 'screenReaderText', array(
+		'expand'   => __( 'expand child menu', 'twentyseventeen' ),
+		'collapse' => __( 'collapse child menu', 'twentyseventeen' ),
```
grappler: It would be good to capitalize the first letter.

functions.php
```diff
@@ -225,7 +220,12 @@ function twentyseventeen_scripts() {
 
 	wp_enqueue_script( 'twentyseventeen-navigation', get_template_directory_uri() . '/assets/js/navigation.js', array(), '20151215', true );
 
-	wp_enqueue_script( 'twentyseventeen-global', get_template_directory_uri() . '/assets/js/global.js', array( 'jquery' ), '20151215', true );
+	wp_localize_script( 'twentyseventeen-navigation', 'screenReaderText', array(
```
grappler: The object name needs to be prefixed.

grappler (commented):
Stating the obvious that this PR will need to be rebased once it is ready to catch up with the changes in master.

comments.php
laurelfulford: I doubled checked with @melchoyce; the reason is to make sure it matches other title case strings in the theme, like 'Leave a Reply'.

components/page/content-front-page-panels.php
laurelfulford: 👍  Thanks!

functions.php
laurelfulford: 👍

functions.php
laurelfulford: Sorry for the delay! I needed to investigate this one a bit - the same lowercase string appears in Twenty Fifteen and Twenty Sixteen, so I wanted to make sure it was okay to replace.

The string was originally set to be lowercase based on this issue: https://core.trac.wordpress.org/ticket/30083#comment:7; in comment 9, it was noted the all-caps actually came from the CSS, not the string's formatting, but the string was left as lowercase just in case.

I double-checked with @davidakennedy and it looks like it'll be ok to capitalize the first word in these strings. That will be more consistent with other strings in the theme, and shouldn't cause issues with how the screen reader reads it.

ntwb:
Would it not be better to merge these changes now, it would make iterating on the code base much more "version control" friendly allowing contributors to iterate on the "current" code base.

There's currently ~24 days of commits here 😱

ntwb:
Thanks for merging 👍

PR #92, opened 2016-09-24
NEW: Normalize Updated to v4.2.0 Fixes #78

### LINKED ISSUE #78: Update normalize.css to the latest version
The latest is v4.2.0 and includes various improvements to what is currently used.

https://github.com/necolas/normalize.css

Keeping all the comments and formatting would also make this easier to maintain.

The other option is to remove it from `style.css` and just include it, this option is currently under consideration for `press-this.css` in [#WP37596](https://core.trac.wordpress.org/ticket/37596)

ahmadawais:
This PR updates normalize to version 4.2.0

[#](https://profiles.wordpress.org/mrahmadawais/) WP.org Username: `mrahmadawais`

### DIFF
--- a/style.css
+++ b/style.css
@@ -40,6 +40,7 @@ Use it to make something cool, have fun, and share what you've learned with othe
 --------------------------------------------------------------*/
 /*--------------------------------------------------------------
 # Normalize v 4.2.0
+@link https://github.com/necolas/normalize.css
 --------------------------------------------------------------*/
 html {
 	font-family: sans-serif;
--- a/style.css
+++ b/style.css
@@ -39,12 +39,13 @@ Use it to make something cool, have fun, and share what you've learned with othe
 	## Galleries
 --------------------------------------------------------------*/
 /*--------------------------------------------------------------
-# Normalize
+# Normalize v 4.2.0
 --------------------------------------------------------------*/
 html {
 	font-family: sans-serif;
-	-webkit-text-size-adjust: 100%;
+	line-height: 1.15;
 	-ms-text-size-adjust: 100%;
+	-webkit-text-size-adjust: 100%;
 }
 
 body {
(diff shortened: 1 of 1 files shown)


josephfusco (changes requested):
We should link back to the normalize repo as someone learning about CSS/WordPress may not understand what it is.

github.com/necolas/normalize.css

ahmadawais:
@josephfusco Done!

ntwb (commented):
I think all the comments from normalize.css should be included. The benefits are two fold, it adds documentation, and it makes it even easier to update future versions in my opinion.

@davidakennedy What do you think?

josephfusco:
@ntwb I agree, but my concerns are that [normalize.css](https://github.com/necolas/normalize.css) uses 2 spaces instead of hard tabs, and the comment structure doesn't entirely match [WordPress CSS coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/#commenting).

ahmadawais:
@ntwb @josephfusco Folks here decided to not have a build process for Twenty Seventeen theme because it can complicate things for beginners. I agreed to that. Which means, everything goes directly into the <kbd>style.css</kbd> file. Adding documentation of Normalize would increase the size of this file beyond something I am comfortable with. I don't think we should do that. 

On that note, I have converted the current CSS from Normalize to use WP Standards.

ntwb:
3rd party libraries _typically_ remain untouched in WordPress, whatever formatting the library uses remains, and is bypassed by JS checks by JSHint etc. We can "ignore" the 2 spaces, and the comment structure because it is a 3rd party library. The documentation value is higher in my opinion than the coding standards.

The `style.css` will no doubt ship with minified versions, this will presumablybe ran manually rather than any build tools in the project, as such those docs will be removed when the minified styles files are created so this also should not be an issue.

grappler:
> The `style.css` will no doubt ship with minified versions

Twenty Sixteen did not include any minified files.

ntwb:
> Twenty Sixteen did not include any minified files.

I did not know that! Ref: https://github.com/WordPress/twentysixteen/issues/77

ahmadawais:
^Yup! That's why it is better for us to not include the docs in there. There ain't anything users would want to modify in a normalize CSS portion.

PR #95, opened 2016-09-25
Update comment styles

tamjidmurtoza:
Some comments have multiline style which I think is not appropriate. For better continuity we should keep them all in same style.

wp.org username: pressionate


davidakennedy:
Thanks for your contribution @tamjidmurtoza!

PR #104, opened 2016-09-27
'edit_post_link' is void functions 

codersantosh:
Also Fix: Undefined variable $type


codersantosh:
Commit ( 918c76b371f37acb15fc4b65237dcb18ebabb770 ) is already requested as pull request.

Issue #105, opened 2016-09-28
Implement preconnect to Google fonts


leobaiano:
Add to preconnect tag to google sources. The matter was discussed in ticket [#37171](https://core.trac.wordpress.org/ticket/37171) and it is already being implemented in the themes that comes with WordPress, then I think it is important to implement this new theme as well.

The implementation should be made in the functions.php through [`wp_resource_hints();`](https://make.wordpress.org/core/2016/07/06/resource-hints-in-4-6/) function; and should only be included if the google fonts is being used, then it is interesting to use [`wp_style_is();`](https://developer.wordpress.org/reference/functions/wp_style_is/) so if a child theme developer decides to remove google preconnect the sources will not be included.

davidakennedy:
@leobaiano Thanks for the issue and PR. Reading through the Core ticket, this seems like a smart thing to do. Let's wait on that since you're working on both. Once that patch is solid, we can make sure everything here is the same and merge.

leobaiano:
OK, thanks very much!

Em 28/09/2016 19:52, "David A. Kennedy" notifications@github.com escreveu:

lukecav:
Would it be worth looking at using Web Font Loader?

https://github.com/typekit/webfontloader

The issue can be if a user does have JavaScript disabled in their browser though.

leobaiano:
@lukecav This should be dealt with in another issue as it has no direct relationship with this one, at least from what I understand. You are suggesting to use the font load to load the font via font-face when the user is with JS disabled in your browser, right? Well, that should be resolved in time to load the source.

leobaiano:
@davidakennedy I believe the ticket [37171](https://core.trac.wordpress.org/ticket/37171) has been resolved. It has to do with the ticket [38121](https://core.trac.wordpress.org/ticket/38121) which has also been resolved.

davidakennedy:
@leobaiano Those aren't merged yet. 😄  Is it better to wait for that?

leobaiano:
ok 😃

PR #106, opened 2016-09-28
Implement preconnect to Google fonts

leobaiano:
Implements preconnect to google sources in accordance with issue #105 

My user in WordPress.org is leobaiano.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -134,6 +134,8 @@ function twentyseventeen_fonts_url() {
 	return esc_url_raw( $fonts_url );
 }
 
+
+
 /**
  * Register widget area.
  *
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -134,6 +134,8 @@ function twentyseventeen_fonts_url() {
 	return esc_url_raw( $fonts_url );
 }
 
+
+
```
grappler: This is not related.

functions.php
```diff
@@ -240,6 +242,24 @@ function twentyseventeen_scripts() {
...
+ * Add preconnect for Google Fonts.
+ *
+ * @since Twenty Seventeen 1.0
+ *
+ * @param array   $urls           URLs to print for resource hints.
+ * @param string  $relation_type  The relation type the URLs are printed.
+ * @return array  $urls           URLs to print for resource hints.
+ */
+function twentyseventeen_resource_hints( $urls, $relation_type ) {
+    // Add preconnect for Google Fonts if queue twentyfourteen-lato.
```
grappler: Please use tabs here instead of spaces.

functions.php
```diff
@@ -240,6 +242,24 @@ function twentyseventeen_scripts() {
...
+	if ( wp_style_is( 'twentyseventeen_fonts_url', 'queue' ) && 'preconnect' === $relation_type )
```
grappler: We need the opening and closing brackets for the if statement.

grappler (changes requested):

functions.php
leobaiano: I had given space to create the function there, when I decided to change place forgot to delete the spaces, thanks for the warning.

leobaiano: I removed the lines.

functions.php
leobaiano: Done.

functions.php
leobaiano: Done.

swissspidy:
Related: https://core.trac.wordpress.org/ticket/37171

leobaiano:
@swissspidy I pointed out that the issue want to open, see #105 😄

davidakennedy:
Refreshed in #491.

PR #113, opened 2016-09-28
Add closing comment tag

bappi:
search.php line 22


search.php
```diff
@@ -19,7 +19,7 @@
 
 			<header class="page-header">
 				<h1 class="page-title"><?php printf( __( 'Search Results for: %s', 'twentyseventeen' ), '<span>' . get_search_query() . '</span>' ); ?></h1>
-			</header>
+			</header><!-- .page-header -->
```
grappler, line 22: I am not sure this is really needed as the tag was just opened two lines above.

bappi: For the continuity of closing tags throughout the theme, it should be better. Besides in `_s` we have this, pretty same [here](https://github.com/Automattic/_s/blob/master/search.php#L20)

PR #117, opened 2016-09-29
Update customizer panel number naming

### LINKED ISSUE #39: Customizer panel number naming
What is the reason to have the panels numbered as `twentyseventeen_panel1` with the numbers sticking to the last work instead of using an underscore to separate the number to make it easier to read e.g. `twentyseventeen_panel_1`?

doughamlin:
Separate panel numbers with underscores. This commit resolves #39.


davidakennedy:
Thanks @doughamlin!

PR #118, opened 2016-09-29
Updated global $twentyseventeencounter comment to refer to correct file

### LINKED ISSUE #31: Remove global $twentyseventeencounter
Remove the `global $twentyseventeencounter;` as it is not being used anywhere. According to the inline comment `//Used in content-frontpage.php` it is used in `content-frontpage.php` but that file does not exist.

I could not find any other references to the global other than in `front-page.php`

doughamlin:
The comment on global $twentyseventeencounter now refers to the correct file. This resolves #31.

### DIFF
--- a/front-page.php
+++ b/front-page.php
@@ -46,7 +46,7 @@ get_header(); ?>
 		$panels = array( '1', '2', '3', '4' );
 		$titles = array();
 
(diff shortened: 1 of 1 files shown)


davidakennedy:
Thanks @doughamlin!

PR #125, opened 2016-09-30
Remove prefixes from theme mod names

### LINKED ISSUE #23: Prefixing theme mod handles
The theme mod handles are prefixed e.g `get_theme_mod( 'twentyseventeen_footer_image' )`.

This has been discussed a few times in the Slack #themereview channel as all of the settings are saved separately for each theme. Prefixing would prevent conflicts with other plugins and core using the setting name.

This decision will affect how other themes set their setting names.

doughamlin:
Removed the "twentyseventeen_" prefix from theme mod names. This resolves #23.

### DIFF
--- a/404.php
+++ b/404.php
@@ -7,6 +7,7 @@
  * @package WordPress
  * @subpackage Twenty_Seventeen
  * @since 1.0
+ * @version 1.0
  */
 
 get_header(); ?>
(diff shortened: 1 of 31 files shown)


doughamlin:
Closing this and submitting a new PR that doesn't mistakenly have three unrelated commits.

PR #139, opened 2016-10-05
Remove extraneous colors

melchoyce:
There are some colors that were inherited from Lodestar, but don’t appear otherwise within the styles. I’ve replaced these colors with colors from the Twenty Seventeen style guide.

### DIFF
--- a/style.css
+++ b/style.css
@@ -128,8 +128,8 @@ dfn {
 }
 
 mark {
-	background-color: #ff0;
-	color: #000;
+	background-color: #eee;
+	color: #222;
 }
 
 small {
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -522,7 +522,7 @@ abbr, acronym {
 }
 
 mark, ins {
-	background: #fff9c0;
+	background: #feee;
```
ntwb: This should be `eee`? i.e. not the 4 character `feee`

ntwb (changes requested):
Found 1 issue

style.css
melchoyce: Good catch! Fixed.

PR #149, opened 2016-10-05
Theme mod handles

### LINKED ISSUE #23: Prefixing theme mod handles

joshcummingsdesign:
Addresses new get_theme_mod calls added recently. If correct, it closes #23 

WordPress.org username is joshcummingsdesign


joshcummingsdesign:
Also, my WordPress.org username is joshcummingsdesign

PR #168, opened 2016-10-06
Solution for too many menu items issue

### LINKED ISSUE #150: Too many menu items causes the theme to be unusable on desktop
When a user has a lot of top level menu items, you no longer can see the content due top the stickyness of the menu.  My example test has 101 top level menu items

screen shot 2016-10-05 at 4 59 15 pm

screen shot 2016-10-05 at 4 59 20 pm

joshcummingsdesign:
Removes fixed class at 20 menu items. Chose 20 as that seems to be around where it breaks down, but open to suggestions on this. Otherwise, closes #150 

WordPress.org username: joshcummingsdesign


josephfusco:
This assumes all 20 menu items total to the same amount of characters. If someone has supercalifragilisticexpialidocious for all menu items It would be beyond 2 rows.

I think while on the right track, we should be checking for menu height, which still wouldn't be entirely accurate because of the different language fonts.

joshcummingsdesign:
Agreed. I think height may be better in this situation.

joshcummingsdesign:
Ok, so I set it so that it does not go to fixed when the nav height is more than half the height of the viewport. The idea being that if the nav is taking up more than half of the viewport, it would be blocking too much of the content.

josephfusco:
IMO it should be a lot less than half. Even allowing 2 rows of items covers more screen than I'm comfortable with but I'll let other weigh in on this too.

This is heading in the right direction though.

joshcummingsdesign:
I changed it to one third window height. Another option is to find the height at two rows and cap it there.

We could also hide it in a drawer when it goes over two rows.

Thoughts?

laurelfulford:
Thanks for this @joshcummingsdesign! It works really nicely. I think the 1/3 might be a bit tall still, but your suggestion of stopping it from sticking at two rows sounds perfect.

joshcummingsdesign:
No longer has position fixed when there are more than two rows of menu items.

joshcummingsdesign:
Noticed one small bug on resize. Will push a fix in just a few.

joshcummingsdesign:
Ok now ready for review :)

josephfusco:
@joshcummingsdesign thank you for doing this! Confirmed the 2 row limit works great and on resize. :)

josephfusco (approved):

Issue #166, opened 2016-10-06
RSS widget could use more breathing room between lines


sixhours:
https://cloudup.com/ivcIxAc1FDX

Readability could be improved with some spacing between the post title, date, and content.

leobaiano:
I did a test here including a margin of 10px between items. But my aesthetic sense is not so determined, see looks better please. 😄 

![screen shot 2016-10-06 at 1 53 56 pm](https://cloud.githubusercontent.com/assets/957257/19162055/ac61348c-8bcc-11e6-8c65-7e42691215e4.png)

enodekciw:
Just submitted #175

This is how it looks with 0.75em margins on .rss-date and 0.75em bottom margin on .rssSummary
![2017-rss](https://cloud.githubusercontent.com/assets/2080743/19169086/8704fb84-8c1b-11e6-8c10-89f9c252e125.jpg)

leobaiano:
How the issues of assignees system here?

davidakennedy:
@leobaiano Welcome! If you see an issue you want to take on, just create a pull request and reference the issue number in it. It's first come, first serve. If you have any questions, you can ask in `core-themes` in the WordPress Slack. Thanks!

melchoyce:
Mocked up a version of the widget that brings it closer to how posts look on static front pages:

![rss widget](https://cloud.githubusercontent.com/assets/2846578/19375495/bcb6143e-91a2-11e6-8242-24235c6e214a.jpg)

@enodekciw if you're interested in coding it up, easiest way to communicate styles is either by sending you the Sketch file, or through Zeplin. Let me know if you want to do Zeplin and DM me your email on WordPress Slack (I'm melchoyce there too) and I'll invite you to the project.

laurelfulford:
Fixed in #175.

PR #175, opened 2016-10-06
RSS widget style

enodekciw:
my take on #166 

@enodekciw on WordPress.org

### DIFF
--- a/style.css
+++ b/style.css
@@ -1984,9 +1984,14 @@ h2.widget-title {
 
 /* RSS Widget */
 
+.widget .rsswidget {
+	display: inline-block;
+	margin-bottom: 0.5em;
+}
+
 .widget .rss-date {
 	display: block;
-	margin: 0.75em 0;
+	margin-bottom: 0.75em;
 }
 
 .widget .rssSummary {
--- a/style.css
+++ b/style.css
@@ -2026,12 +2026,12 @@ h2.widget-title {
 }
 
 .widget_rss .rss-date {
-	margin: 1.5em 0 3em 0;
+	margin: 0.5em 0 1.5em 0;
 	padding: 0;
 }
 
 .widget_rss .rssSummary {
-	margin-bottom: 0.375em;
+	margin-bottom: 0.5em;
 }
 
 /* Contact Info Widget */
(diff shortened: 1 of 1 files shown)


josephfusco:
Thanks for the PR! This looks good.

Instead of having both `margin-top` and `margin-bottom` on `.rss-date`, can you just add `margin-bottom` to the date and `margin-bottom` to `.rsswidget` please? This keeps it consistent with the rest of the project only using bottom margins for type.

josephfusco (approved):

enodekciw:
- Rebased on top of the master branch
- Removed top margin from .rss-date
- Added 0.5em bottom margin to .rsswidget

josephfusco:
@enodekciw This looks good now. Thank you for contributing!

enodekciw:
Ok, I actually think it's still not ready to merge.

Just noticed that `.rsswidget` now looks ugly on hover if title spans over two or more lines (since I had to add `display: inline-block;` to make `margin-bottom` work on `.rsswidget`). Here's an example:

uglyrsswidget

Should I revert it back to how it was before? Imo, `display: inline;` looks way better on `.rsswidget` but that leaves us with `margin-top` on `.rss-date`. Opinions?

melchoyce:
Hold on here for now — I'm going to mock up the RSS specifically to better match static front-page posts.

melchoyce:
Added a mockup here: https://github.com/WordPress/twentyseventeen/issues/166#issuecomment-253705526

enodekciw:
Completely rewrote the CSS for RSS widget. Now it looks like this:
rssv2
Live preview: http://2017.dekciw.lt/

josephfusco:
@enodekciw 

Maybe it's just me but I feel like `.rssSummary` should be `margin-bottom: 0.5em` rather than the current `margin-bottom: 0.375em`. 

cc @melchoyce

melchoyce:
Yeah, let's do `.5em` on `.rssSummary` and also do `margin: 0.5em 0 1.5em 0;` on `.widget_rss .rss-date`.

Thanks @enodekciw!

enodekciw:
Should be all good now? :)

melchoyce:
Looks good to me 👍 

I'll make another issue to address the RSS icon, which we might be able to replace/restyle post-SVG-merge.

Issue #163, opened 2016-10-06
Horizontal scroll on narrow screens in Customizer when using a Static Front Page.


sixhours:
https://cloudup.com/iWgKvz9M5lN

Steps to reproduce:
- Open the theme in the Customizer.
- Set a static front page.
- Narrow screen to 1278px or less.
- Don't scroll down; this only happens when the menu is _not_ stuck to the top.

Using Chrome 53, but it's happening in Firefox as well.

ghost:
There were issues with menu scroll down arrow due to transform rotate property - http://take.ms/XylZ9

I have fixed this issue by adding line height: 1 to genericons. By default - for button tag line height 1.5 is applying & its causing issues to increase the width of scroll down arrow.

Now scrollbar is removed - http://take.ms/fFaN5

melchoyce:
Would this be fixed by #206?

ghost:
@melchoyce Above reported issue can fixed by this PR - https://github.com/WordPress/twentyseventeen/pull/253

Please close this issue as well as PR submitted by me - https://github.com/WordPress/twentyseventeen/pull/179

Thanks for pointing this!

samikeijonen:
I can't reproduce this in #253. But I can see horizontal scroll when I have Query Monitor plugin active.

@sixhours can you double check that this is not the case?

sixhours:
Nope, I don't have that plugin.

samikeijonen:
Did you test it on #253?

melchoyce:
I can't reproduce this. Maybe fixed by #206?

laurelfulford:
I was able to reproduce this in the master branch. It looks like it's resolved in #253 for me.

laurelfulford:
Fixed in #322.

PR #179, opened 2016-10-07
Fixed: Horizontal scroll on narrow screens in Customizer when using a Static Front Page.

ghost:
Please see issue here - https://github.com/WordPress/twentyseventeen/issues/163

There were issues with menu scroll down arrow due to transform rotate property - http://take.ms/XylZ9

I have fixed this issue by adding line height: 1 to genericons. By default - for button tag line height 1.5 is applying & its causing issues to increase the width of scroll down arrow.

Now scrollbar is removed - http://take.ms/fFaN5

WordPress.org Username - pratikchaskar
WordPress.org Profile Link - https://profiles.wordpress.org/pratikchaskar


melchoyce:
Looks like this was fixed by #322. Going to close out, but thank you for your contribution :)

ghost:
You are most welcome @melchoyce :)

PR #181, opened 2016-10-07
align the content with navbar on small screens

akshayvinchurkar:
align the content with navbar on small screens. css tricks shows that this is best practice but this add more 16px padding on both sides without any reason.


grappler:
The code should be deleted and not commented out. My CSS skills are not good enough to review if it should be merged or not.

akshayvinchurkar:
ok i will delete that but when i see it has a reference to css tricks i make decision to comment it out

melchoyce:
Closing this in favor of #182

PR #182, opened 2016-10-07
Delete the rule for align content with top navbar on small screens

akshayvinchurkar:
Delete the rule for align content with top navbar on small screens


melchoyce:
Is this PR replacing #181?

style.css
```diff
@@ -3338,4 +3331,4 @@ article.panel-placeholder {
 	.twentyseventeen-front-page .entry-content blockquote.alignright {
 		margin-right: -20%;
 	}
-}
\ No newline at end of file
+}
```
ntwb, line 3334: Don't delete the newline at the end of the file

ntwb (changes requested):

akshayvinchurkar:
ya first i just comment that code but reviewer says you have to delete the code so i delete that its replacing my first PR

melchoyce:
Thanks, I've gone ahead and deleted the old one.

davidakennedy:
Hey @akshayvinchurkar! Thanks for the pull request. I'm not exactly sure what this fixes, and it looks like this change would effect a lot of the styles in a negative way.

I'm going to close this, but please feel free to create a new issue if the problem persist. A screenshot showing the problem would be helpful.

PR #185, opened 2016-10-07
Update style.css

Swapnildhanrale:
The default WordPress admin bar height is 46px, and we give top 32px to the menubar we need to give this 46px, means when menu is fixed then it will get top 46px and not overlap with the admin bar.

### DIFF
--- a/style.css
+++ b/style.css
@@ -2717,7 +2717,7 @@ article.panel-placeholder {
 
 	.admin-bar .site-navigation-fixed.navigation-top,
 	.admin-bar .site-navigation-hidden.navigation-top {
-		top: 32px;
+		top: 46px;
 	}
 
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -3338,4 +3338,4 @@ article.panel-placeholder {
 	.twentyseventeen-front-page .entry-content blockquote.alignright {
 		margin-right: -20%;
 	}
-}
\ No newline at end of file
+}
```
ntwb, line 3341: The file should continue to have a new blank line at the end of the file

Swapnildhanrale:
Fixed in new branch - [pull request ](https://github.com/WordPress/twentyseventeen/pull/187/) So, Now i'm closing this request.

Issue #16, opened 2016-09-10
Use flex for main content layouts


mor10:
The design is an ideal candidate for the use of `display: flex;` (flexbox) for layouts. This would greatly simplify responsive two/three column layouts while providing reasonable backwards compatibility, and would also make RTL support straight forward via `flex-direction: row-reverse;`.

`display: flex;` gracefully degrades to `display: block;` in older browsers meaning if it is used wisely, the worst case scenario is the user will see a center-aligned single column layout. If this is not acceptable, there is [Flexibility.js](https://github.com/jonathantneal/flexibility).

If this is a path worth exploring, I will put together a proof-of-concept.

mrwweb:
I like this idea. One thing to consider is [potential performance impacts](https://jakearchibald.com/2014/dont-use-flexbox-for-page-layout/). That article is a couple years old, so I don't know where things stand these days, but the overall issues it's highlighting seem worth keeping an eye on if this ticket moves forward.

mor10:
Based on my extensive testing, the performance impact is minimal, and further reduced if the CSS is loaded in critical order.

Somewhat related, Jake also posted this earlier this year: https://jakearchibald.com/2016/link-in-body/

melchoyce:
@mor10: Still interested in Flexbox? Could be really cool here.

mor10:
Absolutely. Now that the core design is in place I'll give it a full pass with flex and see how it plays out.

melchoyce:
Awesome, look forward to it :)

webmandesign:
Just a note here about RTL support: 

You don't have to change `flex-direction` to support RTL languages. It works automatically. https://css-tricks.com/almanac/properties/f/flex-direction/

mor10:
Exactly @webmandesign. That's one of the reasons I'm proposing this. The PR shows how well this works.

mor10:
cc @davidakennedy etc.

Based on the conversation in Slack, I don't think we are ready to move to `flex` at this juncture, so I propose closing this issue and punting the idea of using flex for layout to the next default theme.

On that note, for whomever designs Twenty Eighteen I would recommend starting with flex as the layout module and exploring the design benefits to ensure we get the most out of the feature. Adding it in as an enhancement like what was done here will likely lead to the feature getting punted down the road. 

I leave the actual closing of this issue to the people in charge to ensure continuity.

samikeijonen:
[...]

PR #192, opened 2016-10-07
Flex all layouts - first pass

mor10:
First pass replacing all float-based layout positioning with flex(box) for easier management and cleaner RTL style rules. Addresses #16 

Also includes some minimal CSS and markup cleanup.

Right now all flexed items revert to `display:block` (browser default) when flex is not supported causing a vertical layout. To ensure backward compat to older browsers and IE8/9 we should include Flexibility (https://github.com/jonathantneal/flexibility) with a conditional to test for those older browsers so we don't load an unnecessary JS library for current browsers.


style.css
```diff
@@ -3338,4 +3348,4 @@ article.panel-placeholder {
 	.twentyseventeen-front-page .entry-content blockquote.alignright {
 		margin-right: -20%;
 	}
-}
\ No newline at end of file
+}
```
ntwb: The new blank line at the end of the file shouldn't be removed

style.css
```diff
@@ -3228,6 +3232,7 @@ article.panel-placeholder {
 	.nav-links .nav-next .nav-title:after {
 		color: #222;
 		position: absolute;
+		top: -.2em;
```
ntwb: Expected a leading zero, `top: -0.2em;`, not `top: -.2em;`

ntwb:
@mor10 Could you please add `assets/js/flexibility.js` to both the following files so that this file is excluded from the JSHint and JSCS tests:
• https://github.com/WordPress/twentyseventeen/blob/master/.jscsrc
• https://github.com/WordPress/twentyseventeen/blob/master/.jshintignore

ntwb:
There is also ~15 PHP coding standards errors, you can see the list of these in this Travis CI job:
https://travis-ci.org/WordPress/twentyseventeen/jobs/166493519#L688

mor10:
I have no idea why they are showing up. The only change to the current PR
is the enqueueing of flexibility in functions.php and custom prefix
references in style.css. The errors in Travis appear to be other unrelated
things.

ntwb:
@mor10 Right you are for the PHP, apologies.

The `assets/js/flexibility.js` file still needs to be added to the `.jshintignore` and `.jscsrc` files though so the JSCS and JSHint checks ignore it.

mor10:
Will do, first thing in the morning.

mor10:
@josephfusco What refresh do you want? The conflict seems to be the files @ntwb asked me to add in. Do you want me to take them out again?

josephfusco:
@mor10 Yes I just meant those files which I see you added but it's still conflicting.

This might require a rebase to resolve.

mor10:
This is what happens when PRs take too long to merge. I will rebase and make the necessary updates to get rid of the conflicts. Updated PR will not come until late Thursday at the earliest.

mor10:
PR is refreshed, rebased, and ready to go.

josephfusco (approved):

PR #201, opened 2016-10-07
New Styles for built-in Galleries (username @jordesign)

jordesign:
I've updated the basic styles for default WordPress Galleries. A lot of it was just undoing some wayward styles that were being applied. 

**Before**
screen shot 2016-10-08 at 12 39 53 am

**After**
screen shot 2016-10-08 at 9 10 59 am

I've also collapsed the columns as screen size gets smaller (as anything more than 2col was very squishy

*\* Tablet-ish **
screen shot 2016-10-08 at 9 11 19 am

*\* Mobile **
screen shot 2016-10-08 at 9 11 34 am

*_Captions *_
Gallery Caption styles are being declared in the same rule as normal .wp-caption(s)

*\* Hover Styles **
Simple Hover styles on the image itself
screen shot 2016-10-08 at 9 19 17 am


joyously:
You left the 40px margins on `figure`?

jordesign:
@joyously Yeah - I wasn't sure if `figure` was going to be used in other contexts - so didn't want to change the global style

grappler:
It looking nicer, have you tested the different column sizes like 4 columns?

jordesign:
Working nicely with 4 columns - although it does start to get a bit squeezy in the available space. Anything of 5 columns or more means ridiculously small columns - but that's always going to be the case.

image

joyously:
Your screen shot looks like it is a page because there is no sidebar. On mine, which is a post, it shows the images the size for 4 column, but only two fit across (that's without your mods).

melchoyce:
Looking really good. Can you get rid of the left margin on the left-most column and the right margin on the right-most column so it's full-width?

style.css
```diff
@@ -2324,39 +2324,35 @@ object {
...
-}
+.gallery-item a {
+	display:inline-block;
+	border-bottom:none;
+	background:none;
+	border:10px solid transparent;
 
-.gallery-columns-3 .gallery-item {
-	max-width: 33.33%;
 }
```
josephfusco, line 2254: Add new line after closing `}`

style.css
```diff
@@ -2324,39 +2324,35 @@ object {
...
-
-.gallery-columns-4 .gallery-item {
-	max-width: 25%;
+.gallery-item a:hover {
+	background:none;
+	border:10px solid #ddd;
+	
 }
```
josephfusco, line 2259: Too many new lines between these. There should only be 1.

style.css
```diff
@@ -2324,39 +2324,35 @@ object {
...
-.gallery-columns-3 .gallery-item {
-	max-width: 33.33%;
 }
-
-.gallery-columns-4 .gallery-item {
-	max-width: 25%;
+.gallery-item a:hover {
+	background:none;
+	border:10px solid #ddd;
+
```
josephfusco: Remove extra newline here.

jordesign:
@melchoyce The LHS and RHS padding is tricky as by default the thumbnails are set to be center aligned. This gives the neatest experience across a range of possible uses - particularly where the thumbnails may be of different widths. 

There are 2 approaches we _could_ take, although I don't love either of them:
- scale the images so they are always 100% of the `.gallery-columns-3 .gallery-item` width - but this would mean scaling up the thumbnail/small images (when they are selected) which isn't ideal.
- have the thumbnails left aligned - which would keep things flush-left but have a larger gap on the right hand side. Also could result in some weird alignments when there are images of variable widths.

jordesign:
Thanks @josephfusco - definitely needs a tidy.

style.css
```diff
@@ -2276,7 +2276,7 @@ object {
 	max-width: 100%;
 }
 
-.wp-caption {
+.wp-caption, .gallery-caption {
```
ntwb: New lines for each selector, so this would be:

``` css
.wp-caption, 
.gallery-caption {
```

style.css
```diff
@@ -2324,38 +2324,29 @@ object {
 	text-align: center;
 	vertical-align: top;
 	width: 100%;
+	margin: 0 0 1.5em 0;
+	padding: 0 .5em;
```
ntwb: Use a leading zero for decimal values, so `.5em` becomes `0.5em`

style.css
```diff
@@ -2324,38 +2324,29 @@ object {
...
-}
-
-.gallery-columns-5 .gallery-item {
-	max-width: 20%;
+.gallery-item a {
+	display:inline-block;
+	border-bottom:none;
+	background:none;
+	border:10px solid transparent;
 }
```
ntwb, line 2254: Properties should be followed by a colon and a space e.g. `background: none`, not `background:none`

style.css
```diff
@@ -2324,38 +2324,29 @@ object {
...
 
-.gallery-columns-6 .gallery-item {
-	max-width: 16.66%;
+.gallery-item a:hover {
+	background:none;
+	border:10px solid #ddd;
```
ntwb: Properties should be followed by a colon and a space

style.css
```diff
@@ -2324,38 +2324,29 @@ object {
...
-	max-width: 16.66%;
+.gallery-item a:hover {
+	background:none;
+	border:10px solid #ddd;	
 }
 
-.gallery-columns-7 .gallery-item {
-	max-width: 14.28%;
+.gallery-item a img {
+	display:block;
```
ntwb: Properties should be followed by a colon and a space

ntwb (changes requested):
I've added some more coding standards notes

style.css
```diff
@@ -2324,38 +2324,29 @@ object {
 	text-align: center;
 	vertical-align: top;
 	width: 100%;
+	margin: 0 0 1.5em 0;
```
ntwb: Use shorthand properties, i.e.`margin: 0 0 1.5em;`, not `margin: 0 0 1.5em 0;`

jordesign:
Thanks @ntwb - I'll tidy those up.

melchoyce:
@jordesign I was thinking we'd increase spacing between the images to compensate for the lack of left/right margin — but maybe that would look ridiculous. Maybe we should just left-align.

jordesign:
@melchoyce Part of the problem is that gap is going to vary depending on screen size - so we can't really do it in a bulletproof way.

A quick visual preview of what left aligned would look like.
image

and preview with the hover style (which now appears outside the left margin)
image

melchoyce:
Left aligned looks good 👍

jordesign:
Thumbnails are left aligned now. Works well at larger screens.
It's a little unbalanced on smaller screens (screenshot below) but I think it is a reasonable compromise.

image

style.css
```diff
@@ -2325,38 +2326,29 @@ object {
...
+	display: inline-block;
+	border-bottom: none;
+	background: none;
+	border: 10px solid transparent;
 }
 
-.gallery-columns-7 .gallery-item {
-	max-width: 14.28%;
+.gallery-item a:hover {
+	background:none;
```
ntwb: Expected a single space after the colon

style.css
```diff
@@ -2325,38 +2326,29 @@ object {
 	text-align: center;
 	vertical-align: top;
 	width: 100%;
+	margin: 0 0 1.5em;
+	padding: 0 0.5em;
+	width:50%;
```
ntwb: Expected a single space after the colon

style.css
```diff
@@ -2327,45 +2328,38 @@ object {
 
 .gallery-item {
 	display: inline-block;
-	text-align: center;
+	text-align: left;
 	vertical-align: top;
 	width: 100%;
+	margin: 0 0 1.5em -10px;
+	padding: 0 1em 0 0;
+	width:50%;
```
ntwb: Expected a single space after the colon

ntwb (changes requested):
Added a couple of coding standards issues

jordesign:
Ah sorry about that - missed those ones. Update coming.

melchoyce:
Definitely coming along. I'm a little unsure about the super solid border on hover. All of the other borders are either 1px or 2px, so 10px feels out of place. I could see going up to 4px, but not really beyond that.

Alternately, we could do some sort of overlay on hover, or something similar. (Would be extra great to use CSS filters, since they're GPU accelerated, and figure out an IE and Opera-specific fallback.) What do you think about that, @jordesign?

jordesign:
@melchoyce good point - it does make the thick border out of place. I'll have a think about CSS filters - did you have anything in mind?

melchoyce:
@jordesign Thinking either opacity or brightness.

jordesign:
@melchoyce cool - I was thinking opacity too - maybe a little blur or greyscale with it too

melchoyce:
Unfortunately I think blur (and drop-shadow) _isn't_ GPU accelerated :/

jordesign:
ah ok - can just stick with opacity then. Is filter: opacity; better than just using opacity?

melchoyce:
Yup. IDK why. Browsers are weird.

melchoyce:
(More information on that from this awesome presentation by Will Boyd: https://youtu.be/bEoLCZzWZX8)

josephfusco:
> Unfortunately I think blur (and drop-shadow) isn't GPU accelerated :/

While those properties themselves aren't, if they are on/attached to a layer that is, the animation can still be silky smooth, but it really depends on how the moving parts are coded up. For example in the case of animating a box shadow on hover, you can have a layer behind an element that has a box shadow initially, and manipulate it with opacity. Would give the illusion of the box-shadow being animated without all the painting. :)

But +1 for just a simple opacity transition.

melchoyce:
That is very cool. Thanks for the info @josephfusco.

melchoyce:
Whatever's decided here for hovers should also get applied to regular linked images on hover. They look like this right now:

screen shot 2016-10-10 at 8 50 07 pm

(Not necessarily in this PR, but in general)

jordesign:
@melchoyce agreed - will pop that in a separate PR once we've confirmed the style?

melchoyce:
@jordesign Sounds good!

jordesign:
A side note on left aligning images.... When we have a caption - the caption fills the available space to the next column

image

jordesign:
Proposed style for hover images. :link on left, :hover on right
image

melchoyce:
> A side note on left aligning images.... When we have a caption - the caption fills the available space to the next column

Bummer... just an unavoidable issue with default WordPress galleries?

> Proposed style for hover images. :link on left, :hover on right

Looks good from what I can tell.

jordesign:
re: caption The tricky thing is that the width of the caption takes the full column - regardless of the size of the image. When everything is centered - it is less noticeable as it still feeld balanced.

style.css
```diff
@@ -2331,7 +2331,7 @@ object {
 	text-align: left;
 	vertical-align: top;
 	width: 100%;
-	margin: 0 0 1.5em -10px;
+	margin: 0 0 1.5em 0;
```
ntwb: Use shorthand `margin: 0 0 1.5em;`

style.css
```diff
@@ -2340,16 +2340,21 @@ object {
...
 
 .gallery-item a:hover {
 	background: none;
-	border: 10px solid #ddd;	
+	
 }
 
 .gallery-item a img {
 	display: block;
+	transition: filter .2s ease-in;
```
ntwb: Decimals need a leading zero, `0.2s`, not `.2s`

style.css
```diff
@@ -3310,7 +3315,7 @@ article.panel-placeholder {
 		width: auto;
 	}
 
-	/* Gallery Columns */
+	/* Gallery Columns */
```
ntwb, line 3163: Is this change a switch from a tab to a space? i.e. I hope it's not the other way around from a space to a tab.

style.css
```diff
@@ -3375,4 +3380,5 @@ article.panel-placeholder {
 	.twentyseventeen-front-page .entry-content blockquote.alignright {
 		margin-right: -20%;
 	}
-}
\ No newline at end of file
+}
+0
```
ntwb: Keep the newline but remove the `0`

ntwb (changes requested):
Coding standards fixes

jordesign:
Coding standards fixes in place, and merged to current master

joyously:
> Proposed style for hover images. :link on left, :hover on right

I think that looks awful (as on the other non-gallery images). Couldn't you use a minor scale change instead of opacity?

jordesign:
Hey @joyously - can you elaborate on what makes it awful?
In the context of the change animating on hover - it makes a lot more sense than just a static image.

joyously:
I think changing opacity is awful because it degrades the image. It looks bad, so it is using animation for no good reason. If you use something like scale(1.1), it looks like it zooms in a little on hover instead of occluding the image. And remember, touch devices don't have hover, so it's a subset of users seeing it.

jordesign:
The scaling effect _is_ quite a nice one... :)
@melchoyce do you have a thought on this?

melchoyce:
Reviewing the updates — something about it feels kind of laggy and slow? Playing with it in practice, I'm also not sure about the greyscale. What do you think about this instead?

`.gallery-item a:hover img {
    filter: opacity(60%);
}`

Not a huge fan of scaling overall, after using it in prior themes.

jordesign:
I'm totally happy with that aesthetically - and happy for it to be your call too. PR is updated.

style.css
```diff
@@ -2338,45 +2339,43 @@ object {
...
+	display: inline-block;
+	border-bottom: none;
+	background: none;
 }
 
-.gallery-columns-6 .gallery-item {
-	max-width: 16.66%;
+.gallery-item a:hover {
+	background: none;
+
```
laurelfulford: There's an extra space here - the `}` should just be on the line immediately following `background: none;`.

style.css
```diff
@@ -2609,6 +2608,21 @@ article.panel-placeholder {
...
+	}
+
+	.gallery-columns-3 .gallery-item {
+		max-width: 33.33%;
+	}
+
+	.gallery-columns-4 .gallery-item {
+		max-width: 25%;
+	}
+
```
laurelfulford: Small nitpick: this space can probably go, too - all other `@media` queries finish with the last two `}` on different lines, but without an extra line break between them.

style.css
```diff
@@ -2338,45 +2339,43 @@ object {
 
 .gallery-item {
 	display: inline-block;
-	text-align: center;
+	text-align: left;
 	vertical-align: top;
 	width: 100%;
```
laurelfulford: Looks like the width is declared twice here - can the unneeded one be removed?

laurelfulford:
@jordesign, these are looking really good - thanks for this! 

I found a couple other small nitpicks in the code that I added as notes. 

I also spotted a couple minor things while reviewing - when I use the One column layout, it displays as four columns:

Image

The rtl.css styles should also be updated, just for the `.gallery-item` text alignment. 

That should be it! Thanks again for getting these looking sharp, and if you have any questions about the above just let me know!

jordesign:
Thanks @laurelfulford - I'll get those change happening today

jordesign:
Just pushed up a bunch of updates for:
- Whitespace removal
- Making sure 1 & 2 column grids work
- rtl styles

joyously:
Did those updates include the original fix for the columns? My 4 column gallery is still only showing 2 columns.

jordesign:
Should all be good for 4 columns (working for me locally) unless I've missed a particular use-case

rtl.css
```diff
@@ -257,6 +257,15 @@ input[type="checkbox"] {
 	right: auto;
 }
 
+/* Gallery Styles */
+
+.gallery-item,
+.gallery-caption {
+	text-align:right;
```
ntwb: A space is required after the colon please

rtl.css
```diff
@@ -257,6 +257,15 @@ input[type="checkbox"] {
...
+}
+
+
+
```
ntwb: Only a single blank line here, not 3 blank lines

style.css
```diff
@@ -2338,45 +2339,46 @@ object {
...
-	max-width: 33.33%;
+	margin: 0 0 1.5em;
+	padding: 0 1em 0 0;
+	width: 50%;
 }
 
-.gallery-columns-4 .gallery-item {
-	max-width: 25%;
+.gallery-columns-1 .gallery-item {
+	width:100%;
```
ntwb: A space after the colon

style.css
```diff
@@ -2338,45 +2339,46 @@ object {
...
+.gallery-item a:hover {
+	background: none;
 }
 
-.gallery-columns-7 .gallery-item {
-	max-width: 14.28%;
+.gallery-item a img {
+	display: block;
+	transition: filter 0.2s ease-in;
+	-webkit-backface-visibility: hidden;
```
ntwb, line 2263: Why the webkit vendor prefix `-webkit-backface-visibility` property if you're not using `backface-visibility`?

ntwb (changes requested):
Added some coding standards issues and a question on a vendor prefix addition

joyously:
After you pushed those updates, I downloaded the master zip and overwrote my local install. My 4 column and 3 column galleries still only have 2 columns.

style.css
jordesign: missed this one - sorry - fixed now

jordesign:
@joyously Could you check in the chrome inspector or something and let me know what styles are being applied to the .gallery-item div? Screenshot of the applied styles would be great - help me troubleshoot

joyously:
I was misunderstanding where your changes were committed. I thought you had merged those last ones into master. I downloaded the zip of this branch and I do see your changes, but the thumbnails for my 4 column gallery are the same size as the 3 column gallery. And they overlap a **lot**. They also overlap the sidebar a little bit.  What happened to the rules for sizing the thumbnails?
overlapping-gallery

jordesign:
@joyously there's a universal rule in the stylesheet setting a max width on images
image

not sure why that's not coming through for you - but there's definitely some strange things going on for you there compared to what it should be doing.

rtl.css
jordesign: fixed

rtl.css
jordesign: fixed

style.css
jordesign: fixed

melchoyce:
Testing through git, looks like everything's working as expected. 

The combination of the margin-bottom on the `figure` and on `.gallery` means there's a lot of extra space at the bottom:

screen shot 2016-10-12 at 8 27 27 pm

Anything we can do to tighten that up?

ntwb (approved):

joyously:
@jordesign yes, I have the rule for max-width:100%. That's not the problem. The problem is I don't see any rules for 4 column (25%) and 3 column(33%), so my thumbnails are not getting sized down.

jordesign:
@melchoyce Tightened up the bottom margins so there's no double up on the bottom margin. Also tidied up the margin at the bottom of captions too

melchoyce:
Looks good to me!

jordesign:
Cool - should be good from my end too then.

style.css
```diff
@@ -2293,7 +2293,8 @@ object {
 	max-width: 100%;
 }
 
-.wp-caption {
+.wp-caption,
```
ryelle: Remove the trailing space here

style.css
```diff
@@ -2332,51 +2333,50 @@ object {
...
+	transition: filter 0.2s ease-in;
+	-webkit-backface-visibility: hidden;
+	backface-visibility: hidden;
 }
 
-.gallery-columns-8 .gallery-item {
-	max-width: 12.5%;
+.gallery-item a:hover img {
+	filter: opacity(60%);
+
```
ryelle: Remove this extra line & tab

style.css
```diff
@@ -2332,51 +2333,50 @@ object {
...
-.gallery-columns-8 .gallery-item {
-	max-width: 12.5%;
+.gallery-item a:hover img {
+	filter: opacity(60%);
+	
 }
 
-.gallery-columns-9 .gallery-item {
-	max-width: 11.11%;
+.gallery-columns-2 .gallery-item {
```
ryelle: I think this ruleset should live up near `.gallery-columns-1 .gallery-item` (or that one should be down here, either way they should be together)

style.css
```diff
@@ -2609,6 +2609,28 @@ article.panel-placeholder {
...
+
+	.gallery-columns-2 .gallery-item {
+		max-width: 50%;
+	}
+
+	.gallery-columns-3 .gallery-item {
+		max-width: 33.33%;
+	}
+
+	.gallery-columns-4 .gallery-item {
```
ryelle, line 2527: This is redundant, since `.gallery-item` is already `max-width: 25%`, but might be good to leave for clarity? I'm not sure.

style.css
```diff
@@ -2609,6 +2609,28 @@ article.panel-placeholder {
 		font-size: 16px;
 		font-size: 1.0rem;
 	}
+
+	/* Gallery Columns */
+
```
ryelle: Remove the stray tab here

style.css
```diff
@@ -3314,6 +3336,28 @@ article.panel-placeholder {
 		height: auto;
 		width: auto;
 	}
+
+	/* Gallery Columns */
+
```
ryelle: Remove the stray tab here

style.css
jordesign: Good call

style.css
jordesign: I like the idea of it being there for clarity/completeness if that is allowed?

style.css
jordesign: removed

style.css
jordesign: removed

style.css
jordesign: removed

style.css
jordesign: moved

style.css
jordesign: removed

style.css
```diff
@@ -2322,51 +2323,49 @@ object {
...
+	display: block;
+	transition: filter 0.2s ease-in;
+	-webkit-backface-visibility: hidden;
+	backface-visibility: hidden;
 }
 
-.gallery-columns-9 .gallery-item {
-	max-width: 11.11%;
+.gallery-item a:hover img {
+	filter: opacity(60%);
```
ryelle: One more trailing space to remove

style.css
ryelle: Sounds good, if anyone thinks different we can address it later. 👍

joyously:
@jordesign  I found the problem. There is a rule for `.gallery-item a {display: inline-block;}` that makes the image ignore the max-width on the `figure`. You might not see this problem because probably your test setup is using the default media sizes. I think those are small, so I have my thumbnails set at 200.
I have a 4 column gallery followed by a 3 column gallery.
Here is what I see with your latest changes: a lot of overlap
inline-block-galleries

And with one minor change, I get this: proper thumbnail sizing
inline-galleries

jordesign:
@ryelle trailing space removed - sorry about that.
@joyously - that's so bizarre - I don't get that behaviour at all. But I didn't get any drapbacks from removing that rule either - so it is gone - so hopefully the problem is fixed.

davidakennedy:
Thanks @jordesign! Epic work here. I made a few tweaks to fix up some things in https://github.com/WordPress/twentyseventeen/pull/362/commits/300788d336912d416a926c8b5d5b323c772f506d.

Issue #60, opened 2016-09-11
SVG in place if icon font


mor10:
SVG is the emerging standard for displaying scalable icons, taking over for icon fonts. They are true icons, more acceasible, and allow for more flexibility in terms if style and layout.

This would be a good place to introduce a best-practice example of SVGs in a theme.

The only real argument against SVG at this point is backwards compatibility. This can be solved with Grunticon and other tools.

melchoyce:
+1. @davidakennedy and I have been chatting about this as well and we think it's the right theme to ditch icon fonts. :)

samikeijonen:
I have been using SVG icons in my latest public themes. I'd love to contribute on this one if you want. My starting point have been how `wd_s` theme uses SVGs. And here is [SVG related functions](https://github.com/samikeijonen/checathlon/blob/master/inc/functions-icons.php) that I have used in public themes.

It covers at least these areas:
1. How to load SVG sprite in the footer.
2. Helper function for SVG markup which can be use anywhere we have control of the markup.
3. Using SVG icons in Social menu.
4. Adding arrow icon as a dropdown indicator in menus.

Just let me know if there is something where we can start.

mor10:
Very nice @samikeijonen! I have not tested this, so I'll just ask the question: Does this inject the SVG as an inline element, or reference it?

samikeijonen:
I'm not sure about correct English terms but there are two ways I have injected SVGs.

1) Inline from SVG sprite. `prefix_get_svg` function outputs SVG like this:

```
<svg class="icon icon-github" aria-hidden="true" role="img">
   <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-github"></use
</svg>
```

In this method we include SVG sprite in the [wp_footer](https://github.com/samikeijonen/checathlon/blob/master/inc/functions-icons.php#L13). This is the method I'm using in Checathlon theme.

Note that in the Customizer I use [external reference](https://github.com/samikeijonen/checathlon/blob/master/inc/functions-icons.php#L85), otherwise SVG icons won't show up in the preview. There is a core ticket for this.

2) Use external reference from sprite everywhere. In this method we need to use [svguse.js](https://github.com/Keyamoon/svgxuse) for IE browsers. Also in this method we don't need to include SVG sprite before `</body>` element. Output would be like this:

```
<svg class="icon icon-twitter">
  <use xlink:href="path/to/icons.svg#twitter-icon"></use>
<svg>
```

[...]

Issue #176, opened 2016-10-06
Utilise get_theme_file_uri()


johnbillion:
In WordPress 4.7, the new [`get_theme_file_uri()` function](https://make.wordpress.org/core/2016/09/09/new-functions-hooks-and-behaviour-for-theme-developers-in-wordpress-4-7/) is being introduced. This greatly improves the child theming process.

Unless TwentySeventeen is going to be backwards compatibile with WordPress < 4.7, this function should be used when enqueuing CSS and JS files from the theme.

davidakennedy:
@johnbillion I'm totally for this. It means we'll need a `back-compat.php`, but that's okay. I think default themes should try and show the latest and greatest WordPress has to offer, no matter the feature.

joshcummingsdesign:
Opened PR #287 for this enhancement

PR #206, opened 2016-10-08
Replace Genericons with SVG icons

samikeijonen:
Here is first pass using SVG icons instead of Genericons. It's related to issue #60.

WordPress.org username is `sami.keijonen`.


josephfusco:
IMO this pull request, although titled SVG social menu, should be changing all icons as #60 seems to mean a full replacement.

samikeijonen:
That's true but changing all genericon to SVG icons is a lot more work. I wanted to start with social links menu so we can see how SVG icons works, and what kind of code I'm proposing.

I'd rather create separate PR for other icons if this is the way we're going to go.

josephfusco:
Ok great, as long as we don't ship with a partial implementation that sounds good.

samikeijonen:
Yeah definitely all or nothing.

inc/functions-icons.php
```diff
@@ -0,0 +1,180 @@
...
+ * Add SVG definitions to the footer.
+ */
+function twentyseventeen_include_svg_icons() {
+	
+	// Define SVG sprite file.
+	$svg_icons = get_template_directory() . '/assets/images/svg-icons.svg';
+	
+	// If it exists, include it.
+	if ( file_exists( $svg_icons ) ) {
+		echo file_get_contents( $svg_icons );
```
josephfusco, line 20: Not sure how I feel about `file_get_contents()` within this theme. While not as semantic, seems like it would be more efficient to put the svg data within PHP instead.

josephfusco:
All trailing whitespace should be removed. (Hard to point out on GitHub diffs so here is a screenshot of a few)

screen shot 2016-10-08 at 12 24 03 pm

josephfusco:
I think we should make these functions pluggable so child themes can add/remove svg icons easily. Thoughts?

inc/functions-icons.php
samikeijonen: I personally use `require_once`. For some cases `file_get_contents` would be safer choice as explained in [CSS tricks article](https://css-tricks.com/using-svg/#article-header-id-7).

I'm fine with both.

samikeijonen:
I personally think that pluggable functions should be removed all together.  WordPress template system, hooks and filters is better way to make changes in child theme in my opinion.

We could probably even use `get_template_part` function to load SVG file. But then again I'm not sure if it's a good idea to let child theme overwriting original SVG file.

joyously:
I think pluggable functions are good for some things, for instance, something you want to replace entirely. Filters and hooks don't handle that situation well.

samikeijonen:
> Filters and hooks don't handle that situation well.

Template system and custom functions in child theme will handle that. But I'm fine with pluggable functions also. I just don't think it's a priority at the moment in this PR.

What I'd like to see is feedback on the implementation of using SVG 😉 .

samikeijonen:
Okay. I have updated to PR for removing all the Genericons and replacing them using SVG icons. There must be things that I'm missing. And there are CSS issues. I'm just too tired at the moment to check them 😪 

Couple of notes about Font Awesome icons.
1. Arrows, quote, and reply are very different looking in Font Awesome than in Genericons. Arrows and dropdown icon was so different so that I picked those from Genericons. We could do the same for some other icons also.
2. Path and polldaddy social icons are missing in Font Awesome but are included in Genericons. Therefore I used those. Other option is to leave them out.

inc/functions-icons.php
```diff
@@ -0,0 +1,206 @@
...
+	return $svg;
+
+}
+
+/**
+ * Display an SVG icon.
+ *
+ * @param  array $args Parameters needed to display an SVG.
+ */
+function twentyseventeen_do_svg( $args = array() ) {
```
ryelle, line 106: Is this function necessary? You use `echo twentyseventeen_get_svg` in the rest of the theme.

inc/functions-icons.php
```diff
@@ -0,0 +1,206 @@
...
+		'tumblr.com'      => 'tumblr',
+		'twitch.tv'       => 'twitch',
+		'twitter.com'     => 'twitter',
+		'vimeo.com'       => 'vimeo',
+		'wordpress.org'   => 'wordpress',
+		'wordpress.com'   => 'wordpress',
+		'youtube.com'     => 'youtube',
+	);
+
+	return apply_filters( 'twentyseventeen_social_links_icons', $social_links_icons );
```
ryelle, line 204: If a child theme can't change the SVG source, is there a reason to filter this list?

index.php
```diff
@@ -44,8 +44,8 @@
 				endwhile;
 
 				the_posts_pagination( array(
-					'prev_text' => '<span class="screen-reader-text">' . __( 'Previous', 'twentyseventeen' ) . '</span>',
-					'next_text' => '<span class="screen-reader-text">' . __( 'Next', 'twentyseventeen' ) . '</span>',
+					'prev_text' =>  twentyseventeen_get_svg( $args = array( 'icon' => 'previous' ) ) . '<span class="screen-reader-text">' . __( 'Previous', 'twentyseventeen' ) . '</span>',
```
ryelle, line 47: Is the `$args =` a typo here? (and on the following line)

ryelle: Looks like this is also in
- `index.php` (this file)
- `archive.php`
- `comments.php`
- `single.php`
- `components/post/content.php`
- `inc/template-tags.php`

inc/functions-icons.php
```diff
@@ -0,0 +1,206 @@
...
+		'aria_hidden' => true, // Hide from screen readers.
+	);
+
+	// Parse args.
+	$args = wp_parse_args( $args, $defaults );
+
+	// Set aria hidden.
+	$aria_hidden = '';
+
+	if ( true === $args['aria_hidden'] ) {
```
ryelle, line 64: Rather than relying on whether someone's set `aria_hidden`, can we determine based on whether they've added any labels? So if `title` or `desc` is passed in, likely we don't want this hidden.

ryelle:
I would also opt for filtering rather than pluggable functions. Maybe something as simple as filtering the SVG path? It would need to be carried into `twentyseventeen_get_svg` for the customizer preview, too (so I'm not sure if just using `get_template_path` would work).

Or checking to see if a file exists at that path in `get_stylesheet_directory`, before defaulting to  `get_template_directory`? Then we're not allowing arbitrary URLs, while also giving child-themers a way to override the SVGs.

inc/functions-icons.php
samikeijonen: At the moment you need to set `aria_hidden` to false in that case. I left `title` and `desc` just in case even if don't use them in a theme.

I haven't tested all the `a11y` things (different screen readers with different browsers) when using  `title` and `desc`. Here are couple of articles:
- [Tips for Creating Accessible SVG](https://www.sitepoint.com/tips-accessible-svg/)
- [Accessible SVGs](https://css-tricks.com/accessible-svgs/)

inc/functions-icons.php
samikeijonen: Not necessary. But not sure if child themers would like to use that functions also.

inc/functions-icons.php
samikeijonen: At the moment child theme (or plugins) can't change the SVG source but let's think about that. 

Then again child themes can add their own icons and in that case they need to filter supported social links icons.

index.php
samikeijonen: Do you mean it should just be `twentyseventeen_get_svg( array( 'icon' => 'previous' ) )`?

samikeijonen:
> I would also opt for filtering rather than pluggable functions
> 
> Or checking to see if a file exists at that path in get_stylesheet_directory, before defaulting to get_template_directory?

I like both of these ideas. I'll check them both when I have more time.

Thanks @ryelle all your great feedback!

samikeijonen:
Or like in #176 we could use `get_theme_file_uri`.

josephfusco:
We should add Font Awesome to the copyright section of the project [README.md](https://github.com/WordPress/twentyseventeen/blob/master/README.md#copyright)

samikeijonen:
> We should add Font Awesome to the copyright section of the project README.md

Yes, and still leave the Genericon info since there are couple icons from there.

samikeijonen:
I noticed one big design issue in Quote post format. 

format-quote

This is actually the first item that I don't know how to replace with SVG icon. Genericon is added in `blockquote:before`. And `blockquote` is added in the content.
`

inc/functions-icons.php
ryelle: Right, but I'm suggesting that in the code here you set `aria_hidden` based on whether there's a title/desc -- my thinking is that any time someone sets a title or desc, they will want `aria_hidden` as false, so why not save them a step? And likewise, you want `aria_hidden` to be true if there is no title/desc, because then there's nothing for a screen reader to read.

inc/functions-icons.php
ryelle: Considering we are talking about switching to a child-theme-able setup, I think my comment here doesn't apply - users will be able to add their own SVG file, so they will want to be able to filter this. 👍

index.php
ryelle: Yeah, exactly. You don't reuse `$args`, so there's no reason to set it.

ryelle:
For the quotes - I'm not saying this is the right approach, but we could inject an SVG into the blockquote using the same CSS selector that the quote icon uses. I made a [really quick patch](https://gist.github.com/ryelle/5f06d76cffb08236a91da69abeced8c3) to see how it would work, and it seems to work pretty well. I'm not sure about passing the icon through via `wp_localize_script`, so if there's a better idea I'm for it.

(since this was a quick patch, I also didn't pay much attention to correct styling of the icon)

grappler:
@samikeijonen You will need to rebase before the PR can be merged. The travis check is failing too. Ping me on Slack if you need help to fix it.

samikeijonen:
Okay, now I seemed to fu.. up this PR. Anyway to revert or should I just create new PR.

samikeijonen:
I created fresh PR in #253. Sorry about the mess.

PR #216, opened 2016-10-09
Hide menu and adding print styles

### LINKED ISSUE #214: Missing print styles
Where is the stylesheet for print styles? It looks like it is missing. It should hide the menu, sidebar, etc.

hiddenpearls:
Fixes #214


laurelfulford:
Thanks for this PR, @hiddenpearls!

On reviewing, I noticed a few selectors that don't seem to line up with the theme's styles.

For example, it doesn't look like some of the classes (like `.header-image`, `.content-bottom-widgets`, `.author-bio`, and `.author-title`) are used in the theme. 

There are some selectors, like `.entry-content h4`, that don't like up with the style.css, either - these headings are just referenced like `h4`, `h5`, `h6`, etc. in the style.css.

The print styles also use selectors with post formats (like `.format-image .entry-title`) that aren't used in Twenty Seventeen.

There might be some others that I missed. Can you please review against style.css to make sure everything that's included is needed? Thank you!

melchoyce:
I think we should probably pull all the print styles out into a second stylesheet. Might want to hold off on this until closer to merge.

davidakennedy:
I agree we should wait a bit on merging this.

@melchoyce Why a separate stylesheet? I don't think that's necessary.

hiddenpearls:
@melchoyce a separate stylesheet would create an extra http request.
@laurelfulford I will remove useless selectors tomorrow morning.
@davidakennedy This is going to be my first commit :)

melchoyce:
> Why a separate stylesheet? I don't think that's necessary.

It feels more organized that way, but the second HTTP request is a good point.

davidakennedy:
Merged here: https://github.com/WordPress/twentyseventeen/pull/330

PR #228, opened 2016-10-09
Add a background and border-bottom to featured images.

### LINKED ISSUE #15: Featured image sizes / size limits
The design calls for full-bleed featured images. This poses some interesting challenges:
1. Should the images have a maximum display height? If so, how is increasing width handled in terms of cropping?
2. What happens when the author uploads an image that is narrower than the current viewport width?
3. What happens when the site is displayed on a very wide or zoomed out screen?

Some of this is squarely a design decision, some of it has to do with implementation.

melchoyce:
Fixes #15.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1512,8 +1512,8 @@ body:not(.twentyseventeen-front-page) .entry-header {
 }
 
 .single-featured-image-header {
-    background-color: #fafafa;
-    border-bottom: 1px solid #eee;
+	background-color: #fafafa;
+	border-bottom: 1px solid #eee;
 }
 
 .single-featured-image-header img {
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -1511,6 +1511,11 @@ body:not(.twentyseventeen-front-page) .entry-header {
 	max-width: 740px;
 }
 
+.single-featured-image-header {
+    background-color: #fafafa;
```
grappler: This should be a tab and not spaces

grappler (changes requested):

style.css
melchoyce: The rest of the stylesheet seems to be spaces, though?

grappler: I see something different

image

melchoyce: Weird... I'll copy/paste from existing indents and see if that fixes it.

grappler (approved):

melchoyce:
Thanks @grappler.

Issue #212, opened 2016-10-08
Featured image jiggles on hover in Chrome


josephfusco:
```
macOS 10.12
Google Chrome 53.0.2785.143
```

See gif below for example:

![1ppwwarim9](https://cloud.githubusercontent.com/assets/6676674/19214756/3b8f8cd4-8d59-11e6-883a-728d054d005f.gif)

Issue #196, opened 2016-10-07
Entry-footer and site-footer borders run into each other on posts with comments turned off.


allancole:
This is really minor but it looks like the border-bottom on `.entry-footer` is touching the border-top of the `.site-footer`. 

You can see the issue here: 
![image](https://cloud.githubusercontent.com/assets/709581/19202914/acf866a0-8ca2-11e6-9e86-9a4d702ab071.png)
(Again, it’s subtle so you might have to zoom in to see it 🔍 )

hiddenpearls:
@allancole This actually appears on pages as well when comments form displayed.

jordesign:
Added PR #235 for this issue

PR #235, opened 2016-10-10
Remove double border footer

jordesign:
re: #196 styles to use margin instead of padding on borders on .entry-footer and .site-footer so the border's don't combine to form a 2px section


style.css
```diff
@@ -1730,7 +1731,8 @@ body:not(.twentyseventeen-front-page) .entry-header {
 
 #comments {
 	border-bottom: 1px solid #eee;
-	padding: 2em 0 0.5em;
+	padding: 0 0 0.5em;
+	margin: 2em 0 0 0;
```
ntwb: Use shorthand `margin: 2em 0 0;`, not `margin: 2em 0 0 0;`

ntwb (changes requested):

jordesign:
This PR has become a bit muddied once merged with current master. It seems possibly less needed?

At any rate - it now separates .entry-footer and .site-footer as referenced in #196 
image

style.css
```diff
@@ -3287,6 +3282,7 @@ article.panel-placeholder {
 		font-size: 14px;
 		font-size: 0.875rem;
 		line-height: 1.6;
+		margin-top:3em;
```
ntwb: Expected a single space after the colon

ntwb (changes requested):

melchoyce:
Except for when comments exist on pages, I don't think we actually need the bottom border. I'd rather have the top border on the footer be the divider instead:

screen shot 2016-10-12 at 9 55 41 pm

jordesign:
I reckon even on a page _with_ comments it isn't needed.

Before:
image

After:
image

Only strange outlier seems to be if comments are on and we have post-paging. But I think it still looks alright.
image

style.css
jordesign: this is fixed now

style.css
```diff
@@ -926,7 +926,6 @@ a:hover .nav-title,
 }
 
 .post-thumbnail a img {
-	-webkit-backface-visibility: hidden;
```
laurelfulford, line 929: This was added to fix #212 - is there a reason it's being removed here?

laurelfulford:
This fix looks good @jordesign! Added a question about an unrelated code change, but once that's figured out this should be good to go :)

jordesign:
There was no good reason for that backface visibility being removed - i may have been in the wrong branch. I've popped it back in.

Issue #226, opened 2016-10-09
Default description color hard to read


karmatosed:
Depending on the background you pick the description is a little difficult to read. This actually is the case with the default image provided.

![2016-10-09 at 18 01](https://cloud.githubusercontent.com/assets/253067/19222184/633da0d0-8e4a-11e6-9e57-69158312489f.png)

celloexpressions:
This should probably match the main header text color option; there is enough visual hierarchy between the title and the tagline without the different colors. No need to change it on inner pages currently, although see #189.

jordesign:
In my test I noted that while changing the header color in the customizer changes both the title and tagline colour - neither change is reflected in the front end.

In theory I agree it could probably just match the header color text option. The only variation we might want to consider is giving it an 80% opacity or something like that - so it stands out as _slightly_ different?

jordesign:
On further looking there seems to be 2 issues at play: 
1. The style from the customizer is set to this selector:
   `.site-title a, .site-description`
   which impacts those items on inner pages (where they appear over a background colour) as _well_ as for the homepage if there is a background image. 
2. The selector in style.css
   `.twentyseventeen-front-page:not(.no-header-image) .site-title, .twentyseventeen-front-page:not(.no-header-image) .site-title a`
   has a higher specificity so it is going to overide any user setting from the customizer anyway.

Do we perhaps need to define the homepage header text separately?

jordesign:
Another customizer option like this (but better labeled) would allow us to define overlay title differently?

![image](https://cloud.githubusercontent.com/assets/1842363/19224049/f01b8a48-8ec8-11e6-9557-d0e77b27b521.png)

celloexpressions:
There's no need for another option - the header text color option should only apply to the front page. Specifically, it needs an `active_callback` that returns `( is_front_page() && has_header_image() )`, so it's only shown when it's visible in the preview, and the selectors should be updated for the front page only.

That's probably a discussion for another issue (I think there are several about this), though. It looks like the header text isn't fully implemented yet. This could become the canonical issue for fixing that.

jordesign:
Do we need to allow a control for the header text colour throughout though?
![image](https://cloud.githubusercontent.com/assets/1842363/19224210/e49ec4d8-8ecc-11e6-8002-3aa3dac3b148.png)

celloexpressions:
Definitely not, see #116. The only reason to change this color at all is to make it more readable over an image.

[...]

PR #243, opened 2016-10-11
Swap tagline color for opacity

melchoyce:
Swapping tagline color from #bbb to white (the site title color), and using opacity to make it lighter.

Related to #226.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1041,9 +1041,9 @@ body {
 	margin-bottom: 0;
 }
 
-.twentyseventeen-front-page:not(.no-header-image) .site-description,
-.home.blog:not(.no-header-image) .site-description {
-	color: #bbb;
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -1038,7 +1038,8 @@ body {
 }
 
 .twentyseventeen-front-page:not(.no-header-image) .site-description {
-	color: #bbb;
+	color: #fff;
+    opacity: 0.8;
```
ntwb: Tabs, not spaces

ntwb (changes requested):

style.css
melchoyce: Sorry, I keep pasting in from Inspector. Will fix (along with the merge conflicts)

josephfusco (approved):

PR #246, opened 2016-10-11
Resubmitting new abbr styles

jordesign:
Trying this again

image


melchoyce:
Looks good. To hit color contrast we might have to use `#767676` here — @davidakennedy can you confirm?

jordesign:
I'm totally ok with that change aesthetically - have updated

Issue #240, opened 2016-10-10
Speed up all the animations


melchoyce:
They all feel just little bit slow, IMO.

Maybe cut the animation time in half?

celloexpressions:
I generally find around .15-.18s to feel best, anything faster is hard to tell there's an animation at all. It's tricky here because they're completely inverting the text, so if anything they need to be longer than normal and the slowness could be a result of such a drastic change happening quickly but not instantly. Is there a reason that some properties have different durations than others? That might be contributing to the problem.

If it feels too slow, or IMO, a bit off for some reason, the timing function might be the issue. `ease-in-out` tends to be used a lot, but may not be the most appropriate answer here.

ntwb:
Copying my comment from #247:

In stylelint we have a rule [time-no-imperceptible](http://stylelint.io/user-guide/rules/time-no-imperceptible/):

> _"Disallow `animation` and `transition` less than or equal to 100ms."_

The human eye supposedly can only see animation > 150ms

So, with the PR #247 @josephfusco just submitted changing many instances of `transition` to `100ms` are these transitions actually noticeable at all from applying that PR?

Instead of `100ms` / `0.1s` should these be 150ms? Or removed all together?

melchoyce:
Keeping comments in here for now, rather than #247, in case we spin off into a different direction.

> Instead of 100ms / 0.1s should these be 150ms?

Yeah, makes sense in light of "The human eye supposedly can only see animation > 150ms"

> If it feels too slow, or IMO, a bit off for some reason, the timing function might be the issue. `ease-in-out` tends to be used a lot, but may not be the most appropriate answer here.

I think this might be the case. I feel like I generally have a poor feeling and understanding of animations — I can tell it doesn't feel right, but I'm not sure what right is. Going to ping some folks I know are good at animations to get more feedback and thoughts here.

folletto:
> The human eye supposedly can only see animation > 150ms

Human eye can perceive beyond the standard 60hz (16ms) and up to 200hz-300hz (5-3ms) if I'm not mistaken, [this](http://www.100fps.com/how_many_frames_can_humans_see.htm) is one of many sources you can find on the topic. The issue is not the human eye, but:
- Rendering: some rendering situations just can't render that fast (engine + screen refresh rate + frame rate combo).
- Perception: you _see_ the animation, but you don't _realize_ it's there. The effect however exists. I don't have any study but I'd guess the effect is felt given the brain can process it.

[...]

PR #247, opened 2016-10-11
Cut animation/transition times in half

josephfusco:
Fixes: #240 

wp.org [joefusco](https://profiles.wordpress.org/joefusco)


ntwb:
In stylelint we have a rule [time-no-imperceptible](http://stylelint.io/user-guide/rules/time-no-imperceptible/):

> _"Disallow `animation` and `transition` less than or equal to 100ms."_

The human eye supposedly can only see animation > 150ms

So, with the PR #247 @josephfusco just submitted changing many instances of `transition` to `100ms` are these transitions actually noticeable at all from applying that PR?

Instead of `100ms` / `0.1s` should these be 150ms? Or removed all together?

cc @melchoyce

josephfusco:
@ntwb makes sense.

The fixed navigation sliding out is being removed with #151, but in the places where the color and background color is inverted on hover like the dropdowns and current link styles, IMO it is still noticeable.

ntwb:
Related/Duplicate: #252

josephfusco:
Closing this as #240 discussion is going to change the :hover behavior all together rather than just cutting duration times down.

Issue #250, opened 2016-10-11
Front page edit button styling issue


maheshwaghmare:
**Front Page**
Front page `edit` button not look same as on other page / post edit links.
<img src="https://cldup.com/BRyum8FV97-3000x3000.png" />

---

**Other Page / Post `edit` button**
<img src="https://cldup.com/YUDLYtkShp-3000x3000.png" />

---

**Expected for Front Page**
<img src="https://cldup.com/tUQVjB373T-3000x3000.png" />

melchoyce:
So, I actually think we should remove the edit button in interior pages and instead show "edit" underneath the page title, like we do on the homepage. Posts, since they have meta, can stay the same. cc @laurelfulford

PR #251, opened 2016-10-11
Fixed: Front Page edit link styling issue.

maheshwaghmare:
Fixed: #250 

The _Front Page_ edit button is now looks like this:


melchoyce:
Hi @maheshwaghmare, thanks for your contribution. Reviewing your changes, I don't think this style works well on pages. I know it's inconsistent, but I'm not sure this is the right approach. I'm going to close this PR for now, but please feel free to keep experimenting. Maybe propose some different approaches in #250?

PR #253, opened 2016-10-11
Replace Genericons with SVG icons.

samikeijonen:
Clean PR for replacing Genericons with SVG icons.


melchoyce:
Looks like there's a Travis check failing, @samikeijonen.

samikeijonen:
Yes, I asked @grappler how to solve that. I don't know how that works 😏

josephfusco:
@samikeijonen If you don't have this installed in your text editor I recommend you do. 

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#how-to-use

That contains some info on setting up in a few different editors, but if you use Atom here is an awesome post on that specifically. https://tommcfarlin.com/using-php-codesniffer/

Takes some time setting up but allows you to see all of these inline with your code before they reach travis.

samikeijonen:
Thanks @josephfusco. I have tried Atom but had some weird issues on Windows. Going to check again.

ntwb:
See #60

samikeijonen:
I generate icon in Icomoon app. Should theme include `.json` file so that it's easier to add and remove icons?

This is explained well in [wd_s repo](https://github.com/WebDevStudios/wd_s/tree/master/assets/images/svg-icons#recommended).

carolinan:
I vote no,  it might help developers but there are also users who will not know what a .json file is.

samikeijonen:
True. But how many users would check any of the theme files.

josephfusco:
@samikeijonen maybe we can add it here on GitHub for now and add it to [.gitattributes](https://github.com/WordPress/twentyseventeen/blob/master/.gitattributes) to be excluded from production? I think that is a good compromise in the meantime so it doesn't clutter up the theme and still be available for developers.

gregrickaby:
@samikeijonen I love that you added `aria` support to our `get_svg()` function, if it's cool - I'm going to bring that back into wd_s.

Otherwise, this looks good!

inc/functions-icons.php
```diff
@@ -0,0 +1,203 @@
...
+ *     @type string $title Optional SVG title.
+ *     @type string $desc  Optional SVG description.
+ * }
+ * @return string SVG markup.
+ */
+function twentyseventeen_get_svg( $args = array() ) {
+
+	// Make sure $args are an array.
+	if ( empty( $args ) ) {
+		return esc_html__( 'Please define default parameters in the form of an array.', 'twentyseventeen' );
```
laurelfulford: Escaping on translated strings isn't needed in this theme.

inc/functions-icons.php
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// Define an icon.
+	if ( false === array_key_exists( 'icon', $args ) ) {
+		return esc_html__( 'Please define an SVG icon filename.', 'twentyseventeen' );
```
laurelfulford: Escaping on translated strings isn't needed in this theme.

inc/functions-icons.php
```diff
@@ -0,0 +1,203 @@
...
+
+	// Set ARIA.
+	$aria_labelledby = '';
+
+	if ( $args['title'] && $args['desc'] ) {
+		$aria_labelledby = ' aria-labelledby="title desc"';
+	}
+
+	// Begin SVG markup.
+	$svg = '<svg class="icon icon-' . esc_html( $args['icon'] ) . '"' . $aria_hidden . $aria_labelledby . ' role="img">';
```
laurelfulford: `esc_attr()` is probably a better bet here, since it's being output into the 'class' attribute.

inc/functions-icons.php
```diff
@@ -0,0 +1,203 @@
...
+		'vk.com'          => 'vk',
+		'wordpress.org'   => 'wordpress',
+		'wordpress.com'   => 'wordpress',
+		'yelp.com'        => 'yelp',
+		'youtube.com'     => 'youtube',
+	);
+
+	return apply_filters( 'twentyseventeen_social_links_icons', $social_links_icons );
+
+}
```
laurelfulford, line 203: File should end with a new line.

style.css
```diff
@@ -3245,20 +3091,22 @@ article.panel-placeholder {
...
 		text-align: center;
-		width: 1.5em;
+		width: 2em;
 	}
 
-	.nav-links .nav-previous .nav-title:before {
-		left: -1.5em;
+	.nav-links .nav-previous .nav-title .icon,
+	.nav-links .nav-next .nav-title .icon {
+	}
```
laurelfulford, line 3034: This empty selector should be removed (unless styles should have been added?).

laurelfulford:
These are looking really good, @samikeijonen! 🙂 

I did a visual review and spotted a couple small things:
- Icon for ‘comment by post author’ needs to be added - it was a star originally: https://cloudup.com/caL8E4Us8i0
- There are some things left over in the RTL styles that are causing small issues:
  - There’s a Genericons mention in RTL that's causing double-arrows in the menu: https://cloudup.com/cUEo6rKb0rS
  - Categories and tags icons need to be bumped over: https://cloudup.com/cy1mV_3fumq
  - Previous and next pagination and posts icons: https://cloudup.com/cyi7J0CShGN The RTL styles are adding different CSS content for these areas (left over from Genericons). The arrows should be also rotated, so they still point outwards even though their locations have been swapped.
- You may be looking at other options, but I think @ryelle’s [suggestion for the blockquote icon](https://github.com/WordPress/twentyseventeen/pull/206#issuecomment-252786256) works well - it just needs some minor style tweaks (mostly that `blockquote` has `overflow: hidden`, so the quote icon isn’t visible. Once that's overwritten for quote post blockquotes, it displays nicely).
- I'm sure this is already on your list, but it'll also need an IE8 fallback once everything is good to go.

Just let me know if you have any questions about the above - thanks!

samikeijonen:
Thanks @laurelfulford for review!

I might have not enough time or skills to fix all the issues (IE8 support, adding icons for comment author and blockquote, minor styling issues) before this is merged to Core. I already talked about this with @davidakennedy.

Therefore I suggest that we do final decision about using SVGs or not. If yes, I suggest that we merge this PR. It's helpful on several ways.
- Other people can more easily jump in and fix SVG related issues.
- Avoid double work if there are fixes in Genericon styles in the same time.

inc/functions-icons.php
samikeijonen: Ah, I forgot that Core doesn't use them. Will fix it.

samikeijonen:
@ryelle suggestion seems best for using blockquote icon.

For comment author icon I wish there was `is_post_author()` function. We could then put SVG icon directly in the markup without JS. I'll check how `is_post_author()` function could work.

samikeijonen:
Okay, I messed this PR also. It ain't funny anymore 😄. Everything should be OK but there are other commits seen also in this PR.

Now I need your help how to proceed.
1. Create another clean PR.
2. Merge this if possible.
3. We can always go with Genericons. There have been new commits to that direction also.

PR #254, opened 2016-10-11
Add custom sizes attributes to responsive images

### LINKED ISSUE #120: Wrong picture size as featured image
As a featured image 2000x1200 ist used  as well, but it is resized to 614x369.

A.e 
/uploads/2016/06/IMG_20160320_130911-2000x1200.jpg is resized in HTML or CSS from 2000x1200 to 614x369. Serving a scaled image could save 509.4KiB (90% reduction).

mor10:
Adds control of `sizes` attribute in responsive images to reflect actual displayed size, not only full viewport width. Fixes #120

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -273,7 +273,7 @@ function twentyseventeen_content_image_sizes_attr( $sizes, $size ) {
 	740 <= $width && $sizes = '(max-width: 706px) 89vw, (max-width: 767px) 82vw, 740px';
 
 	if ( is_active_sidebar( 'sidebar-1' ) || is_archive() || is_search() || is_home() || is_page() ) {
-		if ( !( is_page() && 'one-column' === get_theme_mod( 'twentyseventeen_page_options' ) ) ) {
+		if ( ! ( is_page() && 'one-column' === get_theme_mod( 'twentyseventeen_page_options' ) ) ) {
 			767 <= $width && $sizes = '(max-width: 767px) 89vw, (max-width: 1000px) 54vw, (max-width: 1071px) 543px, 580px';
 		}
 	}
--- a/functions.php
+++ b/functions.php
@@ -272,8 +272,10 @@ function twentyseventeen_content_image_sizes_attr( $sizes, $size ) {
 
 	740 <= $width && $sizes = '(max-width: 706px) 89vw, (max-width: 767px) 82vw, 740px';
 
-	if ( is_active_sidebar( 'sidebar-1' ) || is_archive() || is_search() || is_home() ) {
-		767 <= $width && $sizes = '(max-width: 706px) 89vw, (max-width: 767px) 82vw, (max-width: 1000px) 54vw, (max-width: 1071px) 543px, 580px';
+	if ( is_active_sidebar( 'sidebar-1' ) || is_archive() || is_search() || is_home() || is_page() ) {
+		if ( !( is_page() && 'one-column' === get_theme_mod( 'twentyseventeen_page_options' ) ) ) {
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -255,6 +255,54 @@ function twentyseventeen_scripts() {
...
+ *
+ * @param string $sizes A source size value for use in a 'sizes' attribute.
+ * @param array  $size  Image size. Accepts an array of width and height
+ *                      values in pixels (in that order).
+ * @return string A source size value for use in a content image 'sizes' attribute.
+ */
+function twentyseventeen_content_image_sizes_attr( $sizes, $size ) {
+	$width = $size[0];
+
+	//if ( 'page' === get_post_type() || ( ! is_active_sidebar( 'sidebar-1' ) && is_single() ) ) {}
```
laurelfulford: Getting a minor formatting error here - `No space found before comment text; expected "// if (..."`

It looks like it's commented out code - if that's the case, it can just be removed instead. Thanks!

laurelfulford:
Thanks for this PR, @mor10!

I noticed a couple small things I wanted to run by you:

First, I get a bit of a gap for image sizes for the `767px` breakpoint - it's definitely minor, but the `82vw` doesn't seem to cover the whole space all the time:

https://cloudup.com/czv9khWU2q6

What do you think about rolling that one together with the `706px` breakpoint in all cases? Instead of having the two for `706px` and `767px`, it would just be:

`(max-width: 767px) 89vw`

From what I can tell, `89vw` seems to be wide enough to fix the small gap.

Second, when pages are set to use the one column layout in the Customizer, the max image size is 580px:

https://cloudup.com/c-IHFWln_EV

If a check for `'one-column' === get_theme_mod( 'page_options' )` can be incorporated into what you have, with the image size set to `740px` for desktop screen sizes (and smaller for responsive), that should do that trick!

Just let me know if you have any questions at all about the above, and thanks again for this work!

mor10:
@laurelfulford I've updated the PR with the following:
- Fixed the image width issue you pointed out by simplifying the rules.
- Added conditional for page layouts with a stacked logic chain. I tried a whole bunch of different iterations and this seems like the cleanest one.
- Fixed the sizes attribute for featured images on index pages (it was the wrong no-sidebar size in the original PR).

functions.php
```diff
@@ -255,6 +255,55 @@ function twentyseventeen_scripts() {
...
+ * @return string A source size value for use in a content image 'sizes' attribute.
+ */
+function twentyseventeen_content_image_sizes_attr( $sizes, $size ) {
+	$width = $size[0];
+
+	740 <= $width && $sizes = '(max-width: 706px) 89vw, (max-width: 767px) 82vw, 740px';
+
+	if ( is_active_sidebar( 'sidebar-1' ) || is_archive() || is_search() || is_home() || is_page() ) {
+		if ( ! ( is_page() && 'one-column' === get_theme_mod( 'twentyseventeen_page_options' ) ) ) {
+			767 <= $width && $sizes = '(max-width: 767px) 89vw, (max-width: 1000px) 54vw, (max-width: 1071px) 543px, 580px';
```
laurelfulford, line 277: The `theme_mods` had their prefixes removed in #23 - `twentyseventeen_page_options` just needs to be switched to `page_options` and this should be good!

laurelfulford:
Thanks @mor10! Added a comment for one small issue, but that should be it! 🙂 

Also, to verify - is your WordPress.org username mor10 as well?

mor10:
I'll update the prefix. And yes, my username is mor10 pretty much everywhere.

laurelfulford:
Thanks @mor10!

PR #255, opened 2016-10-11
Adding smaller image size

laurelfulford:
Adding smaller custom image size for featured images, for when they're displayed on the blog index and archive pages. Adding function to make sure `$sizes` attribute doesn't exceed viewport. 

See #120.


laurelfulford:
This is a bit redundant and doesn't really address the issue at hand like the existing PR. Going to close and re-submit with a separate issue :)

Issue #203, opened 2016-10-07
Single Line Text Input Height vs Button Height


BharatKaravadra:
When a button is placed next to a text input field, it demonstrates that the button is slightly higher both above and below, than the input field.

This looks a little odd and perhaps the height of buttons could be the same as the text input. See below.

![input button](https://cloud.githubusercontent.com/assets/7373796/19207710/03b1eb72-8ceb-11e6-85b8-41fa9ba79021.png)

PR #262, opened 2016-10-12
Updated button padding with respect to input

sandeshjangam:
Fixes issue [#203](https://github.com/WordPress/twentyseventeen/issues/203).
Button padding changes as per input.

### DIFF
--- a/style.css
+++ b/style.css
@@ -669,7 +669,7 @@ input[type="submit"] {
 	font-size: 0.875rem;
 	font-weight: 800;
 	line-height: 1;
-	padding: 1em 2em;
+	padding: 0.75em 2em;
 	text-shadow: none;
 	-webkit-transition: background 0.2s;
 	transition: background 0.2s;


melchoyce:
Hi @sandeshjangam, thanks for your PR.

Rather than adjusting all of the buttons, what do you think about only changing the padding when buttons are next to inputs?

sandeshjangam:
@melchoyce Definitely we can do this. but IMO buttons size ( Excluding exceptional cases ) should be same throughout the site. Whats your thought on this?

Well, we need to add specific class to button when buttons is next to inputs, so we can change padding  accordingly.

melchoyce:
I think it's okay for buttons in different contexts to be different sizes. What I find more important is that the proportions remain similar — so left/right padding would be adjusted roughly proportionally based on the top/bottom padding change.

ryelle:
> Well, we need to add specific class to button when buttons is next to inputs, so we can change padding accordingly.

You can use sibling selectors, to target any button after an input. [See this article on CSS Tricks](https://css-tricks.com/child-and-sibling-selectors/), specifically "Adjacent sibling combinator `+`" and "General sibling combinator `~`". Which you'll use depends on the markup of these forms.

sandeshjangam:
@melchoyce @ryelle We can add below code with Adjacent sibling combinator. 

`input + button,
input + input[type="button"],
input + input[type="submit"] {
        padding: 0.75em 2em;
 }`

Should I proceed on this?

melchoyce:
@sandeshjangam Please do!

davidakennedy:
Closed in favor of #374.

PR #274, opened 2016-10-13
Account for the sticky menu when scrolling down to the content

ryelle:
When clicking on the "scroll down" arrow in the navigation bar, the scrolled location doesn't account for the sticky menu, and it ends up overlapping content. See:

before
(The post title and search bar are hidden under the sticky menu)

This PR adds the menu's height into the offset, along with a 30px margin (so that menu doesn't sit right on top of the content). This is what it looks like applied:

after

### DIFF
--- a/assets/js/global.js
+++ b/assets/js/global.js
@@ -53,14 +53,15 @@ jQuery( document ).ready( function( $ ) {
 	/**
 	 * 'Scroll Down' arrow in menu area
 	 */
+	$menuTop = -30; // Start with a 30px offset, so the menu doesn't hit the content
 	if ( $( 'body' ).hasClass( 'admin-bar' ) ) {
-		$menuTop = -32;
+		$menuTop -= 32;
 	}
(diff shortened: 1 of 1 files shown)


josephfusco:
This looks good when the front page is latest posts, however I am noticing there is gap when a static front page is set - both with & without admin toolbar.

screen shot 2016-10-12 at 10 36 44 pm

ryelle:
You're right. The `#primary` div starts in different places for each page. On latest posts, there's no space above it, so we need to add that 30px offset. On page-as-front, there's plenty of space, so the 30px adds that gap.

joshcummingsdesign:
#283 Should fix this

melchoyce:
Hey @joshcummingsdesign, I think you're mistaking this PR for an issue?

joshcummingsdesign:
Ah yeah, sorry, been a long day :P

melchoyce:
No worries!

laurelfulford:
Thanks @ryelle!

PR #283, opened 2016-10-13
Addresses menu-scroll-down issue

### LINKED ISSUE #280: menu-scroll-down scrolls too far
The menu scroll down scrolls passed the beginning of the content making it hard to read. It's set to the class '#primary', but maybe we can make it so it scrolls to where the bottom of the nav meets the top of the content.

demo

joshcummingsdesign:
Addresses scroll issue in #280


melchoyce:
Looks like this might be a duplicate of #274?

joshcummingsdesign:
Indeed it is, updated and close duplicate issue.

joshcummingsdesign:
@melchoyce Yes, this addresses the same issue.

assets/js/global.js
```diff
@@ -40,28 +40,27 @@ jQuery( document ).ready( function( $ ) {
...
-		$navigationHeight = $navigation.outerHeight();
-
 		if ( 'none' === $( '.menu-toggle' ).css( 'display' ) ) {
 			$branding.css( 'margin-bottom', $navigationHeight );
 		} else {
 			$branding.css( 'margin-bottom', '0' );
 		}
 	}
 
+	function getNavProps() {
```
tywayne, line 71: I would suggest changing this function name. A 'get' in the function name implies the function will return something. Maybe `setNavProps()`?

tywayne:
Tested this PR, and it appears the offset is different when wpadminbar is present, than when it isn't. We should adjust that offset accordingly.

melchoyce:
Closing this as a duplicate of #274.

PR #286, opened 2016-10-13
Add support for Korean fonts

taggon:
I just added a small amount of code to support Korean fonts. It includes four different fonts because Koreans use different base fonts on each OS.

Related: #59


melchoyce:
Thanks @taggon! 

What do you think about making post titles bold?

screen shot 2016-10-14 at 9 41 33 am

taggon:
@melchoyce My pleasure!

> What do you think about making post titles bold?

It is not necessary but would be good because bold headings are extremely common in Korean websites.

melchoyce:
Sounds good! Would you mind adding bold styles for headings to your PR?

PR #287, opened 2016-10-13
Uses new get_theme_file_uri functions

### LINKED ISSUE #176: Utilise get_theme_file_uri()

joshcummingsdesign:
Uses new get_theme_file_uri(), get_parent_theme_file_uri(), and get_parent_theme_file_path() functions, closes #176

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -17,6 +17,7 @@ if ( version_compare( $GLOBALS['wp_version'], '4.7-alpha', '<' ) ) {
 	require get_template_directory() . '/inc/back-compat.php';
 
 } else {
+
 	/**
 	 * Twenty Seventeen functions and definitions
 	 *
--- a/functions.php
+++ b/functions.php
@@ -9,320 +9,340 @@
  * @since 1.0
  */
 
-if ( ! function_exists( 'twentyseventeen_setup' ) ) :
 /**
- * Sets up theme defaults and registers support for various WordPress features.
- *
- * Note that this function is hooked into the after_setup_theme hook, which
- * runs before the init hook. The init hook is too late for some features, such
- * as indicating support for post thumbnails.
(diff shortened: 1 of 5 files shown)


grappler:
I would not do a `if else` but just do

``` php
/**
 * Twenty Seventeen only works in WordPress 4.7 or later.
 */
if ( version_compare( $GLOBALS['wp_version'], '4.7-alpha', '<' ) ) {
    require get_template_directory() . '/inc/back-compat.php';
}
```

The else is not needed as the code should stop if the `back-compat.php` is called. In `back-compat.php` I would add a new line after the `add_action` so there is a space between the functions.

joshcummingsdesign:
@grappler Without the else statement, PHP throws an error on older versions of WordPress because it will still try to call `get_theme_file_uri()`. Adjusted the spacing.

grappler:
@joshcummingsdesign Then we can use `return` to stop the following code to continue. So

``` php
/**
 * Twenty Seventeen only works in WordPress 4.7 or later.
 */
if ( version_compare( $GLOBALS['wp_version'], '4.7-alpha', '<' ) ) {
    require get_template_directory() . '/inc/back-compat.php';
    return;
}
```

If we are only supporting from 4.7 onwards then we can remove the function `twentyseventeen_the_custom_logo`

davidakennedy:
Excellent work @joshcummingsdesign! Thanks for the feedback @grappler!

Closing in favor of #376.

grappler:
@davidakennedy There is no need to create a new PR see https://github.com/blog/2247-improving-collaboration-with-forks

joshcummingsdesign:
Added return to remove else statement in functions.php. Removed twentyseventeen_the_custom_logo() in favor of the_custom_logo().

Issue #288, opened 2016-10-13
Accessibility-Ready Review: Initial Review


davidakennedy:
I completed an initial accessibility-ready review for the theme. Here's what I found:

**Keyboard Navigation**

Pass: No issues, overall.
- Recommended: Should the menu toggle have a solid border on `:focus`? Right now, it's dotted, like links.

**Controls**

Pass: Pass: No issues, overall.
- Recommended: The button, `<button class="menu-scroll-down"><span class="screen-reader-text">Scroll Down</span></button>` really performs the task of a link, and should probably be changed to an `<a>`.

**Skip Links**

Pass: No issues.

**Forms**

Pass: No issues.

**Headings**

Needs work:
- No `<h1>` exist on the front page when the panel solution is set up. This also happens when the front page is a static page and not posts.

**ARIA Landmarks**

Pass: No issues.

**Link Text**

Needs work:

`link-more` links on search pages need post title to have more context.

**Contrasts**

Needs work:
- `.site-description`: #777 on `#fafafa`. @melchoyce says this can change to `#666666`.
- `.social-navigation` links: `#ffffff` on `#bbb`. @melchoyce says this will change to `#767676` with `:hover` and `:focus` moving to `#333333`.

**Images**

Pass: No issues. Will need to retest when SVGs merged.

**Media**

Pass: No issues.

**Screen Reader Text**

Pass: No issues.

**Disallowed**:
- Any positive tabindex attribute. Negative or zero value tabindex is allowed in specific circumstances (assessed on a case-by-case basis): Pass.
- The inclusion of the accesskey attribute: Pass.
- Spawning new windows or tabs without warning the user: Pass.

**Recommended**:
- Enlarging text: Pass.
- title attributes: Remove title attribute on search form.

davidakennedy:
@laurelfulford One small thing I noticed while testing something else:

This should have screen reader text: https://github.com/WordPress/twentyseventeen/blob/master/inc/template-tags.php#L40

Like this: https://github.com/WordPress/twentyseventeen/blob/master/inc/template-tags.php#L22

For consistency sake.

laurelfulford:
There should be a separate PRs for each of the reported issues and recommendations. 🙂

davidakennedy:
These should all be addressed now.

rianrietveld:
My 2 cents on WordCamp Netherlands:
I love the theme, beautiful!

[...]

PR #290, opened 2016-10-13
Switching menu toggle to use solid border on :focus

laurelfulford:
Changed menu toggle to use a solid border on :focus, as was recommended in #288.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1171,7 +1171,7 @@ body {
 	font-size: 0.875rem;
 	font-weight: 800;
 	line-height: 1.5;
-	margin: 0 auto;
+	margin: 1px auto 2px;
 	padding: 1.0em;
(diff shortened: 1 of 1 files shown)


melchoyce:
Weird request, but: is there any way to force it to be 1px taller, so it loses the optical illusion of it being short being caused by the dark border pushing against the light border? 

screen shot 2016-10-13 at 4 07 18 pm

laurelfulford:
It's worth a shot! 

I tried adding a tiny margin around the button, so you can see the white around the dark outline on `:focus`. Let me know if that does the trick - if no, I can probably 'steal' some of the button's padding to add more space 🙂

melchoyce:
Looks good 👍

josephfusco (approved):

PR #298, opened 2016-10-13
Adding h1 to front page's content panel to improve accessibility

laurelfulford:
Adding h1 to front page's content panel to improve accessibility - takes care the issue with the headers.

See #288.

### DIFF
--- a/components/header/site-branding.php
+++ b/components/header/site-branding.php
@@ -14,7 +14,7 @@
 
 		<?php twentyseventeen_the_custom_logo(); ?>
 
-		<?php if ( is_front_page() && is_home() ) : ?>
+		<?php if ( is_front_page() ) : ?>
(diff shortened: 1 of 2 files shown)


ryelle (approved):

ryelle:
Actually reconsidering "ready to merge" (sorry) -- Is there a reason why the `h1` shouldn't be on the site title on the front page, even when it's panels? I understand doing it for interior pages, since the page title is more important contextually, but on the front page I would think the site title should be first-level info.

davidakennedy:
@ryelle I think you're right here. I was thinking in terms of a normal page, but in context of panels, and `front-page.php` the site title is probably best as `h1`.

@laurelfulford will work on a change.

laurelfulford:
Thanks @ryelle! The static front page should now use an `<h1>` on the site title, and an `<h2>` on the front page's actual page title.

ryelle:
Looks good now :)

Issue #200, opened 2016-10-07
Submenu of leftmost menu item is cut off


joyously:
Hovering on the leftmost menu item shows the submenu, however, that submenu is truncated on the left.
![menu-1st-missing](https://cloud.githubusercontent.com/assets/7957680/19206525/270b4208-8caf-11e6-8395-568cf36c7a00.jpg)

josephfusco:
The same issue exists in Twenty Sixteen (see: https://github.com/WordPress/twentysixteen/issues/394) and was brought up a lot but never resolved. Would be great to include a fix here 👍

joyously:
`.main-navigation ul ul li:hover > ul,
    .main-navigation ul ul li.focus > ul {
        left: 100%;
        right: auto;
    }
`
is the problem. It works with `{left: 0;}`, and probably needs a different fix for RTL.

Additionally, the submenus are quite wide, and it could be a problem on a submenu on the far right. It is set to `width: 15em` which is a lot if you take the number used in the media query (55em, 67em,79em) and divide by 15, that's how many menu items would fit across.

melchoyce:
+1 to fixing this. It's bugged me quite a few times on Twenty Sixteen.

jordesign:
How do we feel about this appearance?
Its tricky to align the little arrowhead on the menu with the down arrow in the top level item, 
so we'd likely need to just left-align the arrowhead to the text above.

![image](https://cloud.githubusercontent.com/assets/1842363/19226509/c4b6783e-8ef7-11e6-8c2c-3b4f2f78ac1d.png)

davidakennedy:
Thanks for the report @joyously!

Seems reasonable @jordesign! PR welcome if you have it.

jordesign:
PR #303 submitted

PR #303, opened 2016-10-14
Make submenu left aligned

jordesign:
This is a fix for issue #200 

Making submenus (dropdowns) positioned relative to the left of the parent LI rather than to the right. This is a simple change of positioning - none of the other styling or functionality has been changed.

Before:
image

After:
image

### DIFF
--- a/style.css
+++ b/style.css
@@ -2784,7 +2784,7 @@ article.panel-placeholder {
 		content: "";
 		height: 0;
 		position: absolute;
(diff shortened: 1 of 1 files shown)


josephfusco:
Should we add anything to rtl.css for this?

jordesign:
Good question re: rtl. My gut feel is that it still works - but open to suggestions

Issue #307, opened 2016-10-14
Make the code for adding panels DRY


davidmosterd:
The panels that are registered in customizer.php generate duplicate code. If the panel registration stays identical, a simple loop removes the risk of one of the panels getting out of sync (e.g. a string with a different punctuation would render a new string to translate).

I will add a PR to show my solution.

davidakennedy:
Replied via comment on the pull request.

PR #308, opened 2016-10-14
Make the code for adding panels DRY

davidmosterd:
username: davidmosterd
issue: #307

### DIFF
--- a/inc/customizer.php
+++ b/inc/customizer.php
@@ -49,7 +49,7 @@ function twentyseventeen_customize_register( $wp_customize ) {
 	) );
 
 	// Panels
-	for( $i = 1; $i <= 4; $i++ ) {
+	for ( $i = 1; $i <= 4; $i++ ) {
 
 		$wp_customize->add_section( 'panel_' . $i, array(
 			'title'           => sprintf( __( 'Panel %s', 'twentyseventeen' ), $i ),
--- a/inc/customizer.php
+++ b/inc/customizer.php
@@ -52,7 +52,7 @@ function twentyseventeen_customize_register( $wp_customize ) {
 	for( $i = 1; $i <= 4; $i++ ) {
 
 		$wp_customize->add_section( 'panel_' . $i, array(
-			'title'           => __( sprtinf( 'Panel %s', $i ), 'twentyseventeen' ),
+			'title'           => __( sprintf( 'Panel %s', $i ), 'twentyseventeen' ),
 			'active_callback' => 'is_front_page',
 			'panel'           => 'options_panel',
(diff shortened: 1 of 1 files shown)


inc/customizer.php
```diff
@@ -52,7 +52,7 @@ function twentyseventeen_customize_register( $wp_customize ) {
 	for( $i = 1; $i <= 4; $i++ ) {
 
 		$wp_customize->add_section( 'panel_' . $i, array(
-			'title'           => __( sprtinf( 'Panel %s', $i ), 'twentyseventeen' ),
+			'title'           => __( sprintf( 'Panel %s', $i ), 'twentyseventeen' ),
```
enodekciw: This is not the way it should be :)
You first have to translate the string, then pass in the variables. Like this:
`sprintf( __( 'Panel %s', 'twentyseventeen' ), $i );`

davidmosterd: You are right, very sorry. I will fix this.

davidakennedy:
@davidmosterd This is a cool idea, and I like the less code. 😄 

But in the case of a default theme, I don't mind doing it the old-fashioned way so anyone learning from the theme understands the Customizer a bit better.

Thank you for the pull request!

davidmosterd:
@davidakennedy Ok, default themes are good for this I agree. I have also considered the option of making this a function (e.g. `twenteseventeen_customize_add_panel`)  and calling it four times with our without parameters. It will not introduce a thing like a for-loop. But I will show people not to repeat code, which is also good? I am new to default-theme development so maybe this is still not the old-fashioned way but wanted to suggest it anyway.

celloexpressions:
I completely disagree with the claim that this is too complex for a default theme. Or even that not doing this is the "old fashioned" way. Using an iterator to loop through and create the multiple instances, still via the vanilla customizer API is the correct way to do this. This is how core would do it if several numbers objects had to be repetitively created. The PR, in my opinion, is exactly how the theme should do this (also eventually making the corrections in #348 to remove the excessive section hierarchy, which will further simplify).

On the contrary, the current approach of copy/pasting the same thing four times adds a lot of excessive complexity, making it look like there's more there than there is because the code is so much longer. This also doesn't need to be moved to a function; a basic for loop in the customize_register callback is the clearest approach.

davidmosterd:
As a programmer: I agree with @celloexpressions (hence the pull request), as we should not learn people the bad practice of repeating yourself and increase the risk of bugs. As a contributor: I'm new to WordPress theme-contribution so I did not/ do not want to sell my opinion too much as I am still learning the culture :)

PR #155, opened 2016-10-06
Update style.css for Japanese Font. Issue #59

nukaga:
Update style.css for Japanese Font. Issue #59

PR #317, opened 2016-10-14
Add support for Japanese fonts

melchoyce:
Adds better support Japanese fonts. Replaces #155. Props @nukaga.

### DIFF
--- a/style.css
+++ b/style.css
@@ -543,6 +543,28 @@ q:after {
 	outline: none;
 }
 
+/* Typography for Japanese Font */
+
+html[lang="ja"] body,
+html[lang="ja"] button,
+html[lang="ja"] input,
+html[lang="ja"] select,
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -543,6 +543,28 @@ q:after {
...
+	letter-spacing: normal;
+}
+
+html[lang="ja"] h1,
+html[lang="ja"] h2,
+html[lang="ja"] h3,
+html[lang="ja"] h4,
+html[lang="ja"] h5,
+html[lang="ja"] h6 {
+	font-family: "Hiragino Kaku Gothic Pro", Meiryo, sans-serif;
```
grappler, line 563: Like we discussed before there is no need to repeat the same CSS for the headers.

melchoyce: Thanks @grappler. Want to submit a PR to fix this?

PR #320, opened 2016-10-14
Revised CSS so that images don't float on smaller mobile devices, eve…

### LINKED ISSUE #316: #page level word-wrap: break-word causes issues
When there are left/right aligned images in a post or page, the `word-wrap: break-word` rule defined on the `#page` ID causes words to be broken down into their individual letters. This is not ideal.

IMO `break-word` should only be applied to titles (if at all).

broken words

mapk:
Revised CSS so that images don't float on smaller mobile devices, even with .alignleft or .alignright class. Media query at 30em allows float to happen on them.

Fixes #316

### DIFF
--- a/style.css
+++ b/style.css
@@ -2325,6 +2325,12 @@ img {
 	max-width: 100%; /* Adhere to container width. */
 }
 
+img.alignleft,
+img.alignright {
+	float: none;
+	margin: 0;
+}
+
 .page-content .wp-smiley,
 .entry-content .wp-smiley,
 .comment-content .wp-smiley {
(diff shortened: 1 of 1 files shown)


joyously:
This assumes that the image is big. What about small images that the user wants to float?

samikeijonen:
> What about small images that the user wants to float?

That's a good question. I'd probably do nothing or center the image.

Kind of the same idea when sidebar is not left or right on narrow screens.

mapk:
> What about small images that the user wants to float?

The image simply displays left-aligned below 30em width. It prevents weirdness in the text. Here's an example.

no-float-2

This is an example of leaving the float and changing to `word-wrap:normal;` on `.entry-content p` which still enables the weird text breaking depending on length of words.

float-2

joyously:
Since the theme is required to style `alignleft` and `alignright`, it should not matter what the screen size is, they should go left and right.

karmatosed:
I think this is a good idea, no images should float on small devices as that causes bad experiences and weirdness with characters. +1 to merge this.

PR #332, opened 2016-10-15
Fixes 314 - Hiding HRs at bottom of post when there is no meta data

### LINKED ISSUE #314: Hide meta borders when meta is empty logged out
If you're logged out and viewing a post without meta, you still see the borders coming from the empty `<footer class="entry-footer">`:

screen shot 2016-10-14 at 12 54 48 pm

You shouldn't be able to see those borders.

mapk:
Fixes #314  - Hiding HRs at bottom of post when there is no meta data


mapk:
Well that stinks. I also just realized that the `edit` link is no longer visible too. I'll have to rework this again.

karmatosed:
@mapk are you able to refresh this?

mapk:
I tried a couple more things and got the edit link back, but without styling b/c I seem to have lost the footer tag. I think someone with better PHP skills would be able to tackle this in a few minutes much better than myself.

laurelfulford:
I added another PR for the issue (#470), so I'm going to close this one out. Thanks for taking a swing at this, @mapk!

Issue #305, opened 2016-10-14
Thumbnail images being scaled larger


jordesign:
In a recent update all images were set to width:inherit; which makes them the width of the parent container. From line 2263...

```
img {
    height: auto; /* Make sure images are scaled correctly. */
    width: inherit;  /* Make images fill their parent's space. Solves IE8. */
    max-width: 100%; /* Adhere to container width. */
}

```

A result of this is that when a thumbnail is inserted - it is scaled up to fill the container (and looks awful in the process).
![image](https://cloud.githubusercontent.com/assets/1842363/19376811/d277e952-922c-11e6-8777-43c23a727df6.png)

Can we revisit this? It strikes me the physical dimensions of the image should be a limit.
Wouldn't max-width:100%; do the same thing?

joyously:
No, you want width:auto, which is the default, but specifying it explicitly fixes IE8.

mapk:
I'm not seeing this issue. My thumbnails are showing up just fine with the `width: inherit;` attribute set. These are images within a post.

![thumbnails](https://cloud.githubusercontent.com/assets/617986/19405596/240a7c16-922e-11e6-9117-4d55fb106d23.png)

jordesign:
I'm noticing it specifically on galleries - where the parent has a width set, so the image inherits it

laurelfulford:
I'm seeing it too, on single posts that have smaller featured images. They're getting stretched to fit the full width of the window.

laurelfulford:
@jordesign I tried moving the IE8 specific style to the IE8 stylesheet in #338 - this fixed the issue I was seeing. Does it fix what you're seeing in the galleries?

jordesign:
Looks good to me - thanks @laurelfulford

laurelfulford:
@jordesign Cool - thanks for checking!

PR #338, opened 2016-10-15
Move IE8-specific image sizing style to ie8.css

laurelfulford:
Related to issue #305.

Moved IE8 specific `img` style from style.css to ie8.css. It seemed to be causing some odd scaling issues in other browsers.


jordesign:
This looks good.

Issue #340, opened 2016-10-15
Potentially misaligned footer text


karmatosed:
I see the following and wonder if the footer link is misaligned?

![2016-10-15 at 14 18](https://cloud.githubusercontent.com/assets/253067/19410369/6fcdadf8-92e2-11e6-9764-1fe45efb0e90.png)

melchoyce:
@karmatosed, can you make a screenshot of the whole page?

karmatosed:
Sure, here you go:

![screencapture-local-wordpress-trunk-dev-1476543293971](https://cloud.githubusercontent.com/assets/253067/19410938/c47b2b20-92ef-11e6-99fb-163c38eeebe5.png)

karmatosed:
Whilst investigating it seems that site-info has a margin -6%.

![2016-10-15 at 15 56](https://cloud.githubusercontent.com/assets/253067/19410956/fa3ff448-92ef-11e6-9194-3fd16329dba8.png)

melchoyce:
Ah, seems to be related to how it's positioned when there isn't a social menu. Thanks :)

yogasukma:
Sorry for not confirmed before, but I create pull request about this before https://github.com/WordPress/twentyseventeen/issues/342 is there anything I need to provided?

melchoyce:
Hi @ysupr, can you add two screenshots: one of the site after your PR is applied without a social menu, and one of the site after your PR is applied with a social menu?

yogasukma:
Ah, it seems it became another issue. 

this is applied PR without social menu
![screen shot 2016-10-15 at 11 25 58 pm](https://cloud.githubusercontent.com/assets/1949957/19411161/ddf1a432-932e-11e6-8091-23e86efc7f1e.png)

and this is after adding social menu
![screen shot 2016-10-15 at 11 27 44 pm](https://cloud.githubusercontent.com/assets/1949957/19411164/f17da38e-932e-11e6-9981-6eeb7091afcb.png)

i tried to revert the theme before applied PR but still, social menu look not good (just adding more margin in left into site info)

yogasukma:
By new issue I mention above are, 
1. After adding the icon, site info seems forced to be centered by adding margin-left 6%. Since it percentage, I am not sure it will be center on every screen width. 
2. And the social icon hit the footer line above it

laurelfulford:
The second one is a known issue, and has been reported here: #304 🙂

PR #342, opened 2016-10-15
fixing misaligned footer text

yogasukma:
fixing for issue #340

this will remove margin on left side of site info.

this is from screenshot before fixing, by @karmatosed 
6fcdadf8-92e2-11e6-9764-1fe45efb0e90

and this is after fixed
screen shot 2016-10-15 at 10 18 02 pm

wp.org : [yogasukma](https://profiles.wordpress.org/yogasukma)


yogasukma:
Nice to mention that the issue are showing while no social media menu.  But when setup social media (even just one) it will looks like centered

ryelle:
Used this as a base for #430 - thanks @ysupr!

PR #349, opened 2016-10-15
Update sticky icon to appear on all post formats, only on blog index

laurelfulford:
The sticky icon was only appearing on post formats using the content.php file - so it was not appearing on sticky audio, video, gallery or image posts:

Image

When it was displaying, it was also appearing on single posts, and whenever the post appeared, not just in its stuck position. Here's an example of the sticky post:

Image

This PR makes sure the icon only appears on the blog index, when the post is 'stuck' (and not in the regular flow), and for all post formats.


samikeijonen:
Looks good.

Issue #323, opened 2016-10-14
Top menu dropdown icon not shown when falling back to wp_page_menu


samikeijonen:
Top menu dropdown icon is not shown when falling back to `wp_page_menu`. This happens when no menu is assigned in `top` menu.

Dropdown arrow is added via `nav_menu_item_title` filter in `inc/functions-icons.php`. I guess `wp_page_menu` is missing that filter.

I'm not sure how to resolve this if there is no similar filter added in the Core.

davidakennedy:
Thanks @samikeijonen! Saw your pull request as well.

I'd like to avoid the JS-only solution if possible. Could the icon not be added with the `'link_after'` parameter in `wp_page_menu`?

samikeijonen:
> Could the icon not be added with the 'link_after' parameter in wp_page_menu?

Yes and I already had that solution. But in there I don't know how to add icon only for page items which have children. We could hide `expand` icon for others via CSS. I'll create separate PR for this solution.

Anyways I'd say that 99.9% of users roll out their own menu.

samikeijonen:
@davidakennedy: I added solution II in PR #380. I personally like this better even it involves little CSS.

PR #350, opened 2016-10-15
Add dropdown icon for page items which has children.

samikeijonen:
Fixes issue #323.

I'm not the biggest fun of adding this icon using JS but it's the easiest and fastest way at this point. Let me know if there is filter which we can use.


davidakennedy:
Closing, we went with #380.

PR #351, opened 2016-10-15
Make sure 'scroll down' arrow is actually hidden on small screens

laurelfulford:
The 'scroll down' arrow is no longer hidden on small screens. Looks like one selector was made more specific, but not the other.

It looked like this:

image

This PR should fix it.


grappler:
There seems to be somethig wrong with this PR.

ntwb:
It looks like the latest commit here in this PR does not want to play nice https://github.com/WordPress/twentyseventeen/pull/351/commits/11ea1d72bec2e3e9743adc2d481b70b276b7fad5

laurelfulford:
Yikes, yeah - will delete and redo.

laurelfulford:
Replaced by #421.

Issue #336, opened 2016-10-15
Accessibility review WCNL


rianrietveld:
My 2 cents on WordCamp Netherlands:
I love the theme, beautiful!

Note: I did not read all the issues in this repo, so maybe some issues are already addresses (sorry for that).

Links:
- outside entry-content there is no hover style on links
- the links in comments have no underline nor style on hover
- menu top level no style on hover

About the entry-meta in an archive, for example:

```
    <span class="posted-on">
        <span class="screen-reader-text">Posted on</span> 
        <a href="http://trunk.dev/2016/09/27/hello-world/" rel="bookmark">
            <time class="entry-date published" datetime="2016-09-27T17:06:23+00:00">September 27, 2016</time>
            <time class="updated" datetime="2016-10-12T11:38:03+00:00">October 12, 2016</time>
        </a>
    </span>
    <span class="byline"> by <span class="author vcard">
        <a class="url fn n" href="http://trunk.dev/author/webbeheer/">webbeheer</a>
    </span>
</span>
```

There is a link with as text the date of a post. This links goes to the post itself, and that link is repeated on the text for the post type. The date is a link to a post, and below that there is also a link to the same post.
On the single page this date is also a link to the age itself
IMHO this date text should not be a link at all, not on the archive and not on the single page.

After that is a link with the name of the author, that goes to the archive of the author.
Including the text "by" to the link is giving the link text more meaning/context .

Also if there are comments the time/date is the comments are also just a link to the date but in fact a link to the comment. Adding more context in the link text might be helpful to screen reader users.

Default header colour (minor issue, but it triggers a validation error)
The site title = #fff
the background = #fafafa
if images are disabled (or still loading) the site-title is not visible
the heading background can be adjusted by the user, but maybe it should be ok out of the box?

Pagination:
Add screen reader text or aria-label to the links to the pages to give more context to the link 
for example:  1 could be Page 1 (page hidden by screen reader text
and also indicate the active page
And add the text _Page_ to the next page arrow link

I think that the scroll down button should stay a button (disagree with https://github.com/laurelfulford/twentyseventeen/commit/bcd5af4d22aedafd97a80be1c97d8de14e132ca7)
the "link" doesn't go to a href, it's just an action to scroll the page down a but, so it should stay a button.

I hope this is helpful

rianrietveld:
About the search form: 
Why choose this form? 

```
<form role="search" method="get" class="search-form" action="http://trunk.dev/">
    <label>
[...]

PR #354, opened 2016-10-15
Add more context to pagination links in archives.

davidakennedy:
- Also fixes missing SVG icons on search page navigation.

See: #336.

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -41,8 +41,9 @@ get_header(); ?>
 			endwhile;
 
 			the_posts_pagination( array(
-				'prev_text' => twentyseventeen_get_svg( array( 'icon' => 'previous' ) ) . '<span class="screen-reader-text">' . __( 'Previous', 'twentyseventeen' ) . '</span>',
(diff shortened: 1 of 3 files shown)


samikeijonen:
Good catch about missing icon in `search.php`!

I have personally put `the_posts_pagination()` function with it's custom args in custom function because it's repeated several times. For example

```
function twentyseventeen_posts_pagination() {

the_posts_pagination( array(
   'prev_text' => twentyseventeen_get_svg( array( 'icon' => 'previous' ) ) . '<span class="screen-reader-text">' . __( 'Previous page', 'twentyseventeen' ) . '</span>',
   'next_text' => '<span class="screen-reader-text">' . __( 'Next page', 'twentyseventeen' ) . '</span>' . twentyseventeen_get_svg( array( 'icon' => 'next' ) ),
   'before_page_number' => '<span class="meta-nav screen-reader-text">' . __( 'Page', 'twentyseventeen' ) . ' </span>',

) );

}
```

This could located for example in `inc/template-tags.php`.

PR #358, opened 2016-10-15
Content variable consistency

### LINKED ISSUE #301: $content variables in content-*.php templates
Working on #258 I've noticed some inconsistency in `content-*.php` templates.

At the moment `$content` variable is defined in:
- `content-video.php`, right after the `.entry-header` (to have it ready for `empty( $video )` check, to fix #258) 
- `content-audio.php`, in `.entry-content` - move it up to make it consistent with `content-video.php`?
- `content-gallery.php`, in `.entry-content` (**never** used) - let's get rid of this one?

enodekciw:
Resolves #301 
- removes `$content` from `content-gallery.php` (which is never used anyway)
- defines `$content` and `$audio` earlier in `content-audio.php` to make it consistent with `content-video.php`

### DIFF
--- a/components/post/content-audio.php
+++ b/components/post/content-audio.php
@@ -31,7 +31,7 @@
 			}
 		?>
 	</header><!-- .entry-header -->
-	
+
 	<?php
 		$content = apply_filters( 'the_content', get_the_content() );
 		$audio = get_media_embedded_in_content( $content, array( 'audio' ) );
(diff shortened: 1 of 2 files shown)


components/post/content-audio.php
```diff
@@ -31,6 +31,11 @@
 			}
 		?>
 	</header><!-- .entry-header -->
+
```
ryelle: Please remove the tab here

enodekciw: Thanks for spotting it, @ryelle

PR #359, opened 2016-10-15
Add support for Chinese fonts

melchoyce:
Add support for zh-HK, zh-TW, and zh-CN. Adds to #59.

**Before**: 

screen shot 2016-10-15 at 3 24 11 pm

**After**:

screen shot 2016-10-15 at 3 24 44 pm

Props @nao and @kaixinlin for input.

### DIFF
--- a/style.css
+++ b/style.css
@@ -545,42 +545,20 @@ q:after {
 
 /* Typography for Chinese Font */
 
-html[lang="zh-HK"] body,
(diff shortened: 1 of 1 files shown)


melchoyce:
Is it possible to target `zh-*` instead of having to use all three?

melchoyce:
The answer appears to be no, but I can do `html[lang^="zh-"]`. :)

PR #360, opened 2016-10-15
Improve search form accessibility.

davidakennedy:
- Add unique `id` to custom search form.
- Move input outside of label.
- This makes search form and label more reliable with assistive technology.

See: #336.


joyously:
Does the Core search work with a different ID than s?

Issue #347, opened 2016-10-15
Add IE8 fallback for SVGs


laurelfulford:
IE8 doesn't support SVGs, so we need to add some fallbacks.

Depending on the icon, we can either replace it with text (by displaying the `screen-reader-text`) or an image fallback.

samikeijonen:
I'd say we could show text at least on social menu. What do you think?

When I generated SVG icons in Icomoon, it also generated PNG files. Only problem with those are that they are always black. Or same color.

I haven't used PNG fallback myself so I need to look into this more deeply.

samikeijonen:
@laurelfulford could you also add `svg` label for all SVG related issues, thanks.

celloexpressions:
I don't think it's worth bundling extra files with the theme for IE8 fallbacks. So let's try to go text-only. The theme should work here, but it doesn't need to look great.

samikeijonen:
> I don't think it's worth bundling extra files with the theme for IE8 fallbacks. So let's try to go text-only. The theme should work here, but it doesn't need to look great.

+10. For some cases text could break the layout or look strange. For example in front page Scroll down text.

For some cases there is no `screen-reader-text` within the icon. Like in sticky post icon. But that's not going to break the theme in IE8 even if there is no icon or any text.

I can create PR for text only fallback.

davidakennedy:
Thanks for the pull request, @samikeijonen!

I'd like to see if using an `<img>` fallback within inline SVG would work. It's [highlighted here](https://css-tricks.com/a-complete-guide-to-svg-fallbacks/).

That way, the theme handles not just IE 8, but any browser that does not support SVGs. It's worth bundling extra images in that case.

samikeijonen:
I can check that tomorrow.

samikeijonen:
PR #381 have basic PNG fallback for SVGs. I'm not the biggest fun of this solution for couple of reasons.
1. PNGs doesn't look very good.
2. It has color contrast issues because it's really hard to maintain PNGs in different colors. See screenshot below.
3. It's not child theme friendly because you can't easily change PNG colors.
4. I don't know are the PNG images loaded even if they are not needed.

![fallback-png](https://cloud.githubusercontent.com/assets/1820415/19416664/5be65aa2-939f-11e6-85c1-73a263f4a3ae.png)

joyously:
Why would you put the background color into the PNG? The image should be transparent and use the theme color for background-color behind it. That's why you use PNG and not JPG.

[...]

PR #361, opened 2016-10-15
Text fallback for SVG icons in IE8.

samikeijonen:
Quick fix for text fallback in IE8. Related to issue #347.

It's a start what I had in mind for social menu.


davidakennedy:
Closing, we went with #347.

PR #367, opened 2016-10-15
Use get_the_title() instead

grappler:
We should use core functions instead of trying to do the same thing ourselves.

WordPress.org username: grapplerulrich


davidakennedy:
Thank you!

PR #369, opened 2016-10-16
Simplify frontpage check for scrollto

grappler:
`is_front_page()` regardless if it is a page or the blog page it will return true. Changing the check to just `is_front_page()` simplifies it.


davidakennedy:
Thanks!

PR #372, opened 2016-10-16
Escape settings on output

grappler:
The settings need to be escaped on output.


davidakennedy:
Thanks @grappler!

celloexpressions:
If we don't trust the filterable saturation that we set, and actually for the hue also, why aren't we escaping these where they're set to variables? The point of that was to make the CSS portions readable. `esc_attr` works against that and adds complexity that is unnecessary - we can call it once at the top of the file for each variable.

Escaping each value 31 times is wasteful in terms of file size and to a lesser extend efficiency and makes the CSS much harder to understand. Why can't we escape once where the variables are set?

celloexpressions:
`esc_attr()` is also too lenient here. Both values should use `absint` as these need to be positive integers, with the saturation having that applied before adding the `%` and converting the number to a string.

PR #374, opened 2016-10-16
Update padding for `<input>` and button elements side by side.

### LINKED ISSUE #203: Single Line Text Input Height vs Button Height

davidakennedy:
Fixes #203.

Props to WordPress.org username: sandesh055 cc: @sandeshjangam

See: #262.

### DIFF
--- a/style.css
+++ b/style.css
@@ -723,6 +723,12 @@ input[type="submit"] {
 	transition: background 0.2s;
 }
 
+input + button,
+input + input[type="button"],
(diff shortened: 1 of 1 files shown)


sandeshjangam:
Thanks @davidakennedy ...

PR #376, opened 2016-10-16
Use new functions available for theme developers in 4.7.

### LINKED ISSUE #176: Utilise get_theme_file_uri()

davidakennedy:
- Also, add `back-compat.php` file. Twenty Seventeen can only be used in 4.7.
- Also uses `the_custom_logo` instead if custom function.
- Cleans up some spacing.

Closes #176.

See #287.


davidakennedy:
Refreshed in #396.

Issue #345, opened 2016-10-15
Smaller logo size


celloexpressions:
I uploaded the WordPress logo that I downloaded straight from .org, and it looked like this:
![twentyseventeen-logo-height-current](https://cloud.githubusercontent.com/assets/7875961/19411924/837f476c-92bf-11e6-9197-b68b5aff525d.PNG)

Most business logos used here probably shouldn't be cropped, but that's the only tool within WordPress for resizing a logo. We should probably set a smaller "recommended" logo height, at least. It's currently 500x500px. Maybe drop the height to 250px? That looks like this for a square logo:
![twentyseventeen-logo-height-250](https://cloud.githubusercontent.com/assets/7875961/19411930/9f13f284-92bf-11e6-8df8-20a4537c6877.PNG)

Note that this is in CSS, but should also be in the custom logo theme support call (I haven't checked the code there).

joyously:
Why would the theme specify any size? It should just be the size that is uploaded.

jordesign:
Do we need to think about allowing for HiDPi/responsive images for the logo as well?

melchoyce:
I think this is a bug — I uploaded a 250px x 250px image and it shows up like this:

<img width="1255" alt="screen shot 2016-10-16 at 1 21 19 pm" src="https://cloud.githubusercontent.com/assets/2846578/19419343/74937608-93a3-11e6-977c-893317d11551.png">

Plus, the CSS specifies a max-width/height of 500px, which is somehow being ignored here:

`@media screen and (min-width: 48em)
.twentyseventeen-front-page:not(.no-header-image) .custom-logo-link img, .home.blog:not(.no-header-image) .custom-logo-link img {
    max-height: 500px;
    max-width: 500px;
}`

melchoyce:
cc @davidakennedy @laurelfulford ^

melchoyce:
> Do we need to think about allowing for HiDPi/responsive images for the logo as well?

I think this is a bug in how core itself handles images in the Customizer?

jordesign:
Would it be overkill to add a range control to let the user set the width in css?

celloexpressions:
The custom logo control will crop the image to the size the user specifies in the customizer, and resize it to the theme's dimensions. The theme should specify the desired size of the logo in the add theme support, as core will try to match that size as closely as possible. Flex width and height allow arbitrary aspect ratios, but one dimension will try to match the width and height numbers in the theme (potentially upsizing images).

[...]

Issue #119, opened 2016-09-29
Update Screenshot.png


jordesign:
I was thinking adding the screenshot would be a _perfect_ first pull request as a new contributor - but then was double checking the guidelines.

'The screenshot should be of the actual theme as it appears with default options'
Is this meant to be with the Theme Unit Test content?

My current appearance using WPTest.io content is this
![image](https://cloud.githubusercontent.com/assets/1842363/18975890/f43adbf4-86f0-11e6-8992-10ea08180b87.png)

josephfusco:
It doesn't need theme unit test content, but we should finish implementing the design first and have proper dummy content in place. :) Good to leave this open though.

jordesign:
Fair Call - 
So this can wait until #4 is finalised and closed

josephfusco:
@melchoyce should we update the default custom header image to the one in the design for this? I can make a separate issue for it.

Currently using:

![header](https://cloud.githubusercontent.com/assets/6676674/19335389/da268fda-90d0-11e6-8289-7ac000e8ba5e.jpg)

jordesign:
Seems it would be worthwhile - have we finalised the default custom content.

melchoyce:
Waiting on #134.

celloexpressions:
Once we have the header image in, we need to decide what the site title and tagline should be, whether there should be a logo, and what menu items should be visible. I think we'll be able to put that content in and take a screenshot directly at 1200x900px and should be good to go. 

We can collaborate on the content shown in the screenshot here. I'd like to see a logo, since this is geared at businesses. I think a smallish white-on-transparent WordPress logo (no text) would make sense.

celloexpressions:
Here's a first pass (at the required 1200x900 size):
![screenshot](https://cloud.githubusercontent.com/assets/7875961/19412185/0a35040e-92c5-11e6-87f1-4fdfc5c50782.png)

We need to decide how big logos should be in #345. I used "Twenty Seventeen" for the site title, and "Bringing your site to life" as the tagline (pulled from the theme description). However, this highlights the difficulty of reading the tagline. I still think it needs either to match the title color without an opacity reduction, or to be made larger.

[...]

Issue #189, opened 2016-10-07
Header on non-front pages


celloexpressions:
Currently the header image (eventually video) is only shown on the front page. However, on inner pages, the header feels a bit plain/empty, especially when a logo is in use:

![twentyseventeen-inner-header](https://cloud.githubusercontent.com/assets/7875961/19192441/1cedd4ba-8c5b-11e6-85d6-ba88fa1765f1.png)

(note: that's also using custom colors from #116)

It feels a bit like wasted space, especially relative to the dramatic size of the featured image below. Could we float the logo when present on larger screens here? Is there a way to either condense the amount of space the header takes, or add visual interest here? It might be nice to have a smaller header image here on inner pages, but it might distract from the featured image below.

davidakennedy:
This would be a good one to have @melchoyce weigh in on.

melchoyce:
I was actually thinking about this when @laurelfulford and I first chatted about logo support, but I'm worried about how super wide logos would look next to the title/tagline. Debating which is the lesser of two evils here and kind of torn, since neither option is particularly great — but displaying a title/tagline is a user choice.

celloexpressions:
If it's floated, the text would drop down if it's too long to fit, but maybe it would also make sense to give the logo a max height and width in this scenario to make it look better?

melchoyce:
Is there some combination of CSS selectors that would let us scale down the logo if there's a title/tagline present, but let it default to its current max-width if the title/tagline are hidden?

paaljoachim:
To me it sounds like there are specifics with header front page and header the rest of the site.
Giving the customizable option for the user to define header media - image or video front page. Then header rest of the site. Having controls that makes it obvious for the user that they first define the front page header and then right below define it for the site in general.

paaljoachim:
Another usage of the non front page header images could be as featured images that show the title of the page/post. Something similar to this:

![full-width-featured-image-behind-title-wordpress](https://cloud.githubusercontent.com/assets/5323259/19257579/a660aeda-8f3e-11e6-9ebc-d683c463a5ea.jpg)

celloexpressions:
@melchoyce mentioned wanting to address this before #119, so it needs to be in the core merge milestone.

celloexpressions:
Superseded by #413.

PR #395, opened 2016-10-16
Scaling down logo and getting it to sit beside site title

laurelfulford:
Related to #345.

Updating logo styles to be similar to mockup in #119. Site title and description will float next to the logo when it's present, and the hope is that the logo will be a bit larger when the site title is hidden.

Currently this behaviour is there, but it uses 'refresh' to update 'header_textcolor' in the Customizer, which isn't great. I haven't actually reduced the image size yet, either.

Would appreciate any thoughts on improving how the Customizer handles this, if possible!

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -47,3 +47,4 @@ Version 1.0:
 @yoavf
 @karmatosed
 @sandesh055
+@adammacias
--- a/assets/css/editor-style.css
+++ b/assets/css/editor-style.css
@@ -1,0 +1,415 @@
+/*
+Theme Name: Twenty Seventeen
+Description: Used to style the TinyMCE editor.
+*/
+
+
+/**
+ * Table of Contents:
+ *
+ * 1.0 - Body
+ * 2.0 - Typography
+ * 3.0 - Elements
+ * 4.0 - Alignment
+ * 5.0 - Caption
+ * 6.0 - Galleries
+ * 7.0 - Media Elements
+ * 8.0 - RTL
+ */
+
+/**
+ * 1.0 - Body
+ */
+
+body {
+	background-color: #fff;
+	color: #333;
+	margin: 20px 40px;
+	max-width: 580px;
+}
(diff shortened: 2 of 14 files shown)


melchoyce:
How do you feel about:

> Is there some combination of CSS selectors that would let us scale down the logo if there's a title/tagline present, but let it default to its current max-width if the title/tagline are hidden?

From https://github.com/WordPress/twentyseventeen/issues/189#issuecomment-252521199?

laurelfulford:
> Is there some combination of CSS selectors that would let us scale down the logo if there's a title/tagline present, but let it default to its current max-width if the title/tagline are hidden?

Geeze, I really overcomplicated this 😐  Will try that instead - thanks @melchoyce!

laurelfulford:
Darn it, the markup doesn't actually change whether or not the site title/description are visible - they're just hidden via CSS. 

Will try to figure out something else, but in the meantime, I've reduced the logo size.

celloexpressions:
How about toggling a body class to show/hide? That's easy to preview with `postMessage`.

laurelfulford:
> How about toggling a body class to show/hide? That's easy to preview with postMessage.

Gah, of course! Thanks @celloexpressions - I'm clearly still overcomplicating this 🙂

laurelfulford:
@melchoyce I've made the logo smaller, but I could use your eyes on how this looks and acts, and on the logo's shape. Let me know if you'd like to see any changes there!

celloexpressions:
In the add theme support, I'm guessing `flex-width` should be true, and possibly not the height. That way a larger logo with a title in the image can take over the space of the site title and tagline when they're hidden. Since we're setting the max height in CSS as up to 200px, that should be the `height`, with the width being whatever (probably 200 to encourage a square aspect ratio) and flex-width, not height, being true.

laurelfulford:
I turned this PR into a bit of a mess, so have replaced with #434. Changes actually related to the logo have all made their way over there 🙂

PR #396, opened 2016-10-16
Use new functions available for theme developers in 4.7.

### LINKED ISSUE #176: Utilise get_theme_file_uri()

davidakennedy:
- Also, add back-compat.php file. Twenty Seventeen can only be used in 4.7.
- Also uses the_custom_logo instead if custom function.

Closes #176.

See #287. Refresh of #376.

### DIFF
--- a/components/header/site-branding.php
+++ b/components/header/site-branding.php
@@ -12,7 +12,7 @@
 <div class="site-branding">
 	<div class="wrap">
 
-		<?php twentyseventeen_the_custom_logo(); ?>
+		<?php the_custom_logo(); ?>
 
 		<?php if ( is_front_page() ) : ?>
 			<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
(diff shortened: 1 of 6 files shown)


inc/icon-functions.php
```diff
@@ -13,7 +13,7 @@
 function twentyseventeen_include_svg_icons() {
 
 	// Define SVG sprite file.
-	$svg_icons = get_template_directory() . '/assets/images/svg-icons.svg';
+	$svg_icons = get_parent_theme_file_path( '/assets/images/svg-icons.svg' );
```
johnbillion, line 16: Should this have been `get_theme_file_path()`?

inc/icon-functions.php
```diff
@@ -87,7 +87,7 @@ function twentyseventeen_get_svg( $args = array() ) {
 
 	// Use absolute path in the Customizer so that icons show up in there.
 	if ( is_customize_preview() ) {
-		$svg .= '<use xlink:href="' . get_template_directory_uri() . '/assets/images/svg-icons.svg' . '#icon-' . esc_html( $args['icon'] ) . '"></use>';
+		$svg .= '<use xlink:href="' . get_parent_theme_file_uri( '/assets/images/svg-icons.svg#icon-' . esc_html( $args['icon'] ) ) . '"></use>';
```
johnbillion, line 90: Should this have been `get_theme_file_uri()`?

johnbillion: In fact, the `#icon-*` hash shouldn't be contained within this call.

PR #403, opened 2016-10-16
Add support for Devanagari

melchoyce:
Adds support for `bn-BD`, `hi-IN`, and `mr-IN`. Please let me know if I missed any, or if there’s a better font alternative than Arial available on either Mac or Windows. 

**Before**:

screen shot 2016-10-16 at 5 40 52 pm

**After**:

screen shot 2016-10-16 at 5 40 00 pm

### DIFF
--- a/style.css
+++ b/style.css
@@ -550,6 +550,48 @@ q:after {
 	outline: none;
 }
 
+/* Typography for Devanagari Font */
+
+html[lang="bn-BD"] body,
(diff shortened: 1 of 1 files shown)


grappler:
Nepali uses the same script https://en.wikipedia.org/wiki/Nepali_language

melchoyce:
Thanks @grappler, can you make a new issue for that?

PR #405, opened 2016-10-16
SVG Fallbacks

laurelfulford:
See #347.

Adding fallbacks for the SVGs that have no text labels or other alternatives.

For the pagination previous and next links, social icons, and search button, the `screen-reader-text` has been made visible.

For the mobile menu toggles (also necessary to open/close the menu in IE8), the `^` is inserted as CSS content instead, to preserve space, and is flipped depending on orientation.

Also added another argument to the `twentyseventeen_get_svg()` function, so you can pass 'fallback' HTML to the SVG (needed for situations like the dropdown toggles). 

The styles have all been added to style.css, but they can also be moved to ie8.css.

### DIFF
--- a/assets/css/ie8.css
+++ b/assets/css/ie8.css
@@ -42,6 +42,7 @@ h6 {
 }
 
 /* Main Navigation */
+
 .menu-toggle {
 	width: 150px;
 }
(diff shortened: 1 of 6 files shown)


samikeijonen:
+10 for this approach.

davidakennedy:
@laurelfulford After some testing and code reviewing, I'd like to continue this approach, but use Modernizr's method. It's more battle-tested. 😄 

So:

```
    /**
     * Test if inline SVGs are supported.
     * @link https://github.com/Modernizr/Modernizr/
     */
    function supportsInlineSVG() {
        var div = document.createElement( 'div' );
        div.innerHTML = '<svg/>';
        return ( typeof SVGRect != 'undefined' && div.firstChild && div.firstChild.namespaceURI ) == 'http://www.w3.org/2000/svg';
    }
```

Then:

```
supportsInlineSVG();
        if ( supportsInlineSVG() === true ) {
            document.documentElement.className = document.documentElement.className.replace( /(\s*)no-svg(\s*)/,'$1svg$2' );
        }
```

And add `no-svg` to HTML element.

PR #411, opened 2016-10-17
Style dark media playlists

melchoyce:
Continuation of #407.

Has some issues — for example, when you hover over a playlist item, the background colors seem to change at different speeds. Definitely need some help on this one.

**Before**:

screen shot 2016-10-16 at 11 03 29 pm
screen shot 2016-10-16 at 11 03 22 pm

**After**:

screen shot 2016-10-16 at 11 02 33 pm
screen shot 2016-10-16 at 11 02 46 pm

### DIFF
--- a/style.css
+++ b/style.css
@@ -2284,6 +2284,78 @@ object {
 	border: 0;
 }
 
+/* Playlist Color Overrides: Light */
+
+.site-content .wp-playlist-light {
+	border-color: #eee;
+	color: #222;
+}
+
+.site-content .wp-playlist-light .wp-playlist-current-item .wp-playlist-item-album {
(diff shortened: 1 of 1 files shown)


laurelfulford:
@melchoyce To fix the hover speed issue, you can take this like:

```
 .site-content .wp-playlist-dark .wp-playlist-caption:hover {
       background-color: #aaa;
 }
```

... and change it to:

```
 .site-content .wp-playlist-dark .wp-playlist-caption:hover {
       background-color: transparent;
 }
```

The issue is that there's two 'competing' hover styles happening at the same time - this takes the one that doesn't span the full width of the area out of the running.

PR #412, opened 2016-10-17
Switch to a proper callback function for menu `fallback_cb`

ryelle:
Currently the `fallback_cb` code is always running, displaying a second menu even if one is already set. This should be a proper callback, so I've created a new function for the menu fallback & swapped it into the callback parameter.


samikeijonen:
> displaying a second menu even if one is already set.

I can't reproduce this but it doesn't hurt either.

PR #414, opened 2016-10-17
Add the custom header image to the interior pages.

### LINKED ISSUE #413: Simplify header between front-page and interior pages
After watching a couple conversations unfold across issues, and my own testing, I’ve come to realize that removing the header image on interior pages is a mistake. 

I think we should still show a collapsed header on interior pages, but include a sliver of the header image, along with the gradient overlay. This simplifies a number of issues:
- It fixes #225, and the eliminates any confusion around the header text color picker.
- It makes it easier to add a logo and have it look good on your homepage and your interior pages. Currently a lighter logo might look good on the homepage, but a dark logo looks better on interior pages. Now, the contrast will be the same so the same logo looks good in both locations.

@ryelle is working on a PR. (#414)

ryelle:
Fixes #413

I've refactored the header-image code to display the custom header on all pages, but kept the functionality of only showing a page's featured image on the front page.

I also added a new class to the body, `has-header-image`. None of the existing interior styles have been removed, as those still apply if no header is set. Instead, I'm using the new class to set the site-title color etc, instead of relying on home or `twentyseventeen-front-page`.

The JS is also tweaked slightly to account for interior menu positioning.

I did not touch the other color schemes, but they'll need to be updated too — I can do this tomorrow.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1562,6 +1562,26 @@ body:not(.title-tagline-hidden) .site-branding-text {
 	background-position: center center;
 	background-repeat: no-repeat;
 	background-size: cover;
+	height: 165px;
+}
+
+.custom-header-image:before {
+	/* Permalink - use to edit and share this gradient: http://colorzilla.com/gradient-editor/#000000+0,000000+100&0+0,0.3+75 */
+	background: -moz-linear-gradient(to top, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.3) 75%, rgba(0, 0, 0, 0.3) 100%); /* FF3.6-15 */
+	background: -webkit-linear-gradient(to top, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.3) 75%, rgba(0, 0, 0, 0.3) 100%); /* Chrome10-25,Safari5.1-6 */
+	background: linear-gradient(to bottom, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.3) 75%, rgba(0, 0, 0, 0.3) 100%); /* W3C, IE10+, FF16+, Chrome26+, Opera12+, Safari7+ */
+	filter: progid:DXImageTransform.Microsoft.gradient(startColorstr="#00000000", endColorstr="#4d000000", GradientType=0); /* IE6-9 */
+	bottom: 0;
+	content: "";
+	display: block;
+	height: 100%;
+	left: 0;
+	position: absolute;
+	right: 0;
+}
+
+.twentyseventeen-front-page.has-header-image .custom-header-image,
+.home.blog.has-header-image .custom-header-image {
 	height: 0;
 	padding-top: 66%;
 }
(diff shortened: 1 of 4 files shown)


ryelle:
The color schemes shouldn't need updating since we don't change anything on top of the images. This should be ready for review 😄

style.css
```diff
@@ -1524,6 +1520,26 @@ body {
 	background-position: center center;
 	background-repeat: no-repeat;
 	background-size: cover;
+	height: 165px;
+}
+
+.custom-header-image:before {
+	/* Permalink - use to edit and share this gradient: http://colorzilla.com/gradient-editor/#000000+0,000000+100&0+0,0.3+75 */
+	background: -moz-linear-gradient(top,  rgba(0,0,0,0) 0%, rgba(0,0,0,0.3) 75%, rgba(0,0,0,0.3) 100%); /* FF3.6-15 */
```
ntwb: 1. Gradient should use `to top`, not `top` (A common mistake (matching outdated non-standard syntax) is to use just a side-or-corner without the preceding `to` ([src](http://stylelint.io/user-guide/rules/function-linear-gradient-no-nonstandard-direction/))
2. A single space after each comma is expected

style.css
```diff
@@ -1524,6 +1520,26 @@ body {
...
+	background: -webkit-linear-gradient(top,  rgba(0,0,0,0) 0%,rgba(0,0,0,0.3) 75%,rgba(0,0,0,0.3) 100%); /* Chrome10-25,Safari5.1-6 */
```
ntwb: 1. Gradient should use `to top`, not `top` (A common mistake (matching outdated non-standard syntax) is to use just a side-or-corner without the preceding `to` ([src](http://stylelint.io/user-guide/rules/function-linear-gradient-no-nonstandard-direction/))
2. A single space after each comma is expected

style.css
```diff
@@ -1524,6 +1520,26 @@ body {
...
+	background: linear-gradient(to bottom,  rgba(0,0,0,0) 0%,rgba(0,0,0,0.3) 75%,rgba(0,0,0,0.3) 100%); /* W3C, IE10+, FF16+, Chrome26+, Opera12+, Safari7+ */
```
ntwb: A single space after each comma is expected

style.css
```diff
@@ -1524,6 +1520,26 @@ body {
...
+	filter: progid:DXImageTransform.Microsoft.gradient( startColorstr='#00000000', endColorstr='#4d000000',GradientType=0 ); /* IE6-9 */
```
ntwb: 1. A single space after each comma is expected
2. Expected double quotes, not single quotes, i.e. `"`, not `'`

ntwb (changes requested):
Add CSS Coding Standards review

style.css
ryelle: Considering this is code copied from the existing CSS (and there's another gradient using the same syntax on `.panel-image:before`), I'd rather not make those changes here -- is there another issue or PR for CSS cleanup?

ryelle: Just saw this was fixed with another PR, so it should be good now, right?

ntwb (approved):

melchoyce:
Being finished up in #474.

PR #419, opened 2016-10-17
Use pluggable functions for template functions

grappler:
We should not use pluggable functions if they are being used in hooks as we can use `remove_action/filter` in the child theme. Template files are the only ones that we cannot remove and need to be overriden.

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -83,7 +83,7 @@ function twentyseventeen_entry_footer() {
 endif;
 
 
-if ( ! function_exists( 'twentyseventeen_entry_footer' ) ) :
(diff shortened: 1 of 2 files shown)


inc/template-tags.php
```diff
@@ -83,6 +83,7 @@ function twentyseventeen_entry_footer() {
 endif;
 
 
+if ( ! function_exists( 'twentyseventeen_entry_footer' ) ) :
```
ryelle: This should be `twentyseventeen_edit_link`

ryelle (changes requested):
I like this idea. I think you have a copy-paste error though.

ryelle (approved):

PR #421, opened 2016-10-17
Making selector more specific to hide 'scroll-down' arrow on small sc…

laurelfulford:
Redo of #351: 

The 'scroll down' arrow is no longer hidden on small screens. Looks like one selector was made more specific, but not the other.

It looked like this:

image

This PR should fix it.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1163,7 +1163,7 @@ a:hover .nav-title,
 
 /* Scroll down arrow */
 
-.menu-scroll-down {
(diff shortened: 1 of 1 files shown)


laurelfulford:
Also added a fix for the scroll arrow _not_ displaying when it should - one of the menu icon selectors just needed to be a bit more specific. It should have been hiding the dropdown toggle icons in specific situations, but instead was hiding all icons.

PR #427, opened 2016-10-17
Adjust spacing around social menu based on feedback

laurelfulford:
Adjust the space around the social menu, so it's even on the top and bottom.

Also bump down the site info a bit, so it lines up vertically.

### DIFF
--- a/assets/css/editor-style.css
+++ b/assets/css/editor-style.css
@@ -391,12 +391,6 @@ object {
 	display: block;
 }
 
-.tiled-gallery a,
-.tiled-gallery a:focus,
-.tiled-gallery a:hover {
-	box-shadow: none;
-}
-
 /**
  * 7.0 - Media Elements
  */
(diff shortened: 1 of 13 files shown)


melchoyce:
I think this may have introduced some weirdness at smaller screen sizes: 

**Phone**
screen shot 2016-10-17 at 4 00 48 pm

**Tablet**
screen shot 2016-10-17 at 4 00 53 pm

**Around 960px**
screen shot 2016-10-17 at 4 01 01 pm

laurelfulford:
Yeesh, I was too concerned with the top and bottom, I didn't look closely at the left and right.

Thanks for catching this, @melchoyce! Should be better spaced, now.

melchoyce:
Soooo close! Found one more bug:

screen shot 2016-10-17 at 6 27 18 pm

laurelfulford:
Replacing this with #447 - I made a bit of a mess with the commits.

PR #447, opened 2016-10-18
Fixing spacing around, and above and below social navigation icons

laurelfulford:
Fixing spacing around, and above and below social navigation icons

Also added RTL styles for the 'site-info' alone issue. Replacement for #427.


melchoyce:
Looks good to me!

Issue #425, opened 2016-10-17
Let people make empty menus


melchoyce:
If I'm making a one-page website, I'll often unassign the menu so it disappears. Sometimes this works in themes (woohoo!) and sometimes it doesn't (sad).

I'm thinking that if a menu isn't assigned, we should hide the menubar, rather than showing all of your site's pages by default.

@laurelfulford / @davidakennedy what are your thoughts here?

laurelfulford:
Ah, that's a good point about a one page website not necessarily needing a menu. I'd be fine with there being no fallback menu, to make it easier for users to leave one out.

slaFFik:
There is a possibility to do that. But the end result looks ugly.
Create a menu, uncheck "Automatically add new top-level pages to this menu" checkbox, assign it to "Top" menu location and DO NOT add pages/links/whatever to that menu.

Here is how it will look like:

![image](https://cloud.githubusercontent.com/assets/282234/19453308/5e05b1c4-94bd-11e6-8e69-898e23ab9919.png)

celloexpressions:
The menu fallback should probably be removed, and the styling should also be updated to handle an empty menu as the above screenshot shows.

melchoyce:
Ideally that would be fixed, so the whole bar would disappear. :)

samikeijonen:
+1. I'd say in 99.9% of sites there is need for custom menu or not menu at all.

enodekciw:
If you decide to get rid of wp_nav_menu fallback, PR #458 should resolve this.

davidakennedy:
I'm for this, as I think it allows more flexibility for users. I just want to make sure nothing is missed. Will review the pull request.

laurelfulford:
Merged #458 to fix - thanks @enodekciw!

PR #458, opened 2016-10-18
Remove wp_nav_menu fallback

enodekciw:
#425
- Sets wp_nav_menu fallback_cb to false
- Removes twentyseventeen_fallback_menu(), which isn't needed anymore
- Checks whether the 'top' theme_location has a menu assigned or not. If not - prevents any navigation menu markup from showing up.

### DIFF
--- a/assets/js/global.js
+++ b/assets/js/global.js
@@ -134,7 +134,7 @@
 	$( document ).ready( function() {
 
 		// Let's fire some JavaScript!
-		if( twentyseventeenScreenReaderText.has_navigation === 'true' ) {
+		if ( 'true' === twentyseventeenScreenReaderText.has_navigation ) {
 
 			/**
 			 * 'Scroll Down' arrow in menu area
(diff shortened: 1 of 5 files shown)


samikeijonen:
Couple of notes:
1. When using `has_nav_menu` there is no need for setting fallback at all. We don't get that far.
2. Now there is also no need for adding dropdown icon via JS. There is also couple of lines CSS which can be removed.

On my mobile so harder to give more details and line numbers.

enodekciw:
@samikeijonen 
1. Default wp_nav_menu() fallback_cb is set to 'wp_page_menu'. So we **do need** to set it to false, so that wp_nav_menu doesn't fallback to 'wp_page_menu'. See codex: [Set to false for no callback](https://developer.wordpress.org/reference/functions/wp_nav_menu/#parameters).
2. Good point. I'll dig into this right now, since I realised that now we dont need to enqueue `/assets/js/navigation.js` if no menu is assigned.

samikeijonen:
If you use `has_nav_menu` around nav markup there is no nav at all unless you set one. Therefore `wp_nav_menu` won't run. But it doesn't hurt to have fallback either.

enodekciw:
Oh, now it makes sense. Guess my brain is still sleeping. Sorry :)

melchoyce:
This is looking really good! 

I'm wondering if there’s any way we could make the scrollability more obvious, maybe by adding a down arrow over the header image so there’s some indication that you can scroll more. Or not taking up the whole viewport, so a little bit of the page background peeks through at the bottom.

I'm also totally cool with merging this and then exploring enhancements like that in a new PR. :)

laurelfulford:
@melchoyce, moved your suggestion to #472 🙂

PR #460, opened 2016-10-18
Remove letter-spacing for all non-latin alphabets

### LINKED ISSUE #365: Double-check letter-spacing on non-latin alphabets
Related to #59. Will do once that issue is closed.

melchoyce:
Fixes #365.

### DIFF
--- a/style.css
+++ b/style.css
@@ -857,6 +857,29 @@ html[lang="th"] textarea {
 	line-height: 1.8;
 }
 
+/* Remove letter-spacing for all non-latin alphabets */
+
+html[lang="ar"] *,
+html[lang="ary"] *,
+html[lang="azb"] *,
+html[lang="haz"] *,
+html[lang="ps"] *,
+html[lang^="zh-"] *,
+html[lang="bg-BG"] *,
+html[lang="ru-RU"] *,
+html[lang="uk"] *,
+html[lang="bn-BD"] *,
+html[lang="hi-IN"] *,
+html[lang="mr-IN"] *,
+html[lang="el"] *,
+html[lang="gu-IN"] *,
+html[lang="he-IL"] *,
+html[lang="ja"] *,
+html[lang="ko-KR"] *,
+html[lang="th"] * {
+	letter-spacing: normal;
+}
+
 /*--------------------------------------------------------------
 # Forms
 --------------------------------------------------------------*/
--- a/style.css
+++ b/style.css
@@ -631,7 +631,6 @@ html[lang="ps"] h4,
 html[lang="ps"] h5,
 html[lang="ps"] h6 {
 	font-weight: bold;
-	letter-spacing: normal;
 }
 
 /* Typography for Chinese Font */
@@ -650,7 +649,6 @@ html[lang^="zh-"] h3,
 html[lang^="zh-"] h4,
 html[lang^="zh-"] h5,
 html[lang^="zh-"] h6 {
-	letter-spacing: normal;
 	font-weight: 700;
 }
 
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -857,6 +857,29 @@ html[lang="th"] textarea {
...
+html[lang="bn-BD"] *,
+html[lang="hi-IN"] *,
+html[lang="mr-IN"] *,
+html[lang="el"] *,
+html[lang="gu-IN"] *,
+html[lang="he-IL"] *,
+html[lang="ja"] *,
+html[lang="ko-KR"] *,
+html[lang="th"] * {
+	letter-spacing: normal !important;
```
josephfusco: This seems aggressive. Is `!important` absolutely needed?

melchoyce: I'd love to find an alternative — do you have suggestions?

josephfusco: I have gone through style.css and gathered a list of selectors specifying letter-spacing greater than 0 (normal).

``` css
h5,
.site-title,
.page .panel-content .entry-title,
.page-title,
body.page:not(.twentyseventeen-front-page) .entry-title,
.entry-meta,
.entry-footer .cat-links,
.entry-footer .tags-links,
.nav-subtitle,
.comment-metadata,
h2.widget-title,
.widget_rss .rss-date,
.widget_rss li cite,
.site-content .wp-playlist-current-item .wp-playlist-item-artist,
.twentyseventeen-customizer.twentyseventeen-front-page .twentyseventeen-panel .twentyseventeen-panel-title
```

So looking at these, this PR does reset letter-spacing with the least amount of CSS, however any plugin/child themes that need to change letter-spacing will have a difficult time. Not sure how often that is or if it is even something to worry about.

An alternative to this is to go through each of these selectors above and see if we really need to be using letter-spacing, which would make this PR unnecessary.

melchoyce: In latin alphabets, the letter spacing creates more breathability in the typeface. In non-latin alphabets, the spacing doesn't work as well, especially alphabets like Devanagari or Arabic that blend letters together like ligatures.

When I first looked at this issue, I tried your approach of finding all of the places where letter-spacing occurs and got more than halfway through before realizing just how ridiculous the amount of CSS was. Using `*` seems much more practical. 

If there's a way I can make the list more specific without using `!important`, that would be great. If there isn't, I think we should go with it.

josephfusco (approved):

style.css
josephfusco: Yeah in this case then I would condone its use.

laurelfulford:
@melchoyce If we're doing a sweeping `!important` letter-spacing for all the languages we've included styles for, we probably don't need them on a language by language basis on the headers, too. What do you think about removing those?

laurelfulford:
And we should get the editor-style.css to match whatever's changed here, too - thanks!

josephfusco:
Also we should also be able to use `letter-spacing: 0` instead of `letter-spacing: normal`.
I don't see anything specifically on it in the WP CSS Coding Standards? @ntwb

melchoyce:
> If we're doing a sweeping !important letter-spacing for all the languages we've included styles for, we probably don't need them on a language by language basis on the headers, too. What do you think about removing those?

Makes sense! Done :)

> Also we should also be able to use letter-spacing: 0 instead of letter-spacing: normal.
> I don't see anything specifically on it in the WP CSS Coding Standards? 

Also done :)

melchoyce:
> And we should get the editor-style.css to match whatever's changed here, too - thanks!

@laurelfulford I'll do a second PR for that 👍

melchoyce:
Couldn't get the conflicts to happen locally, so I opened a new PR: #473

joyously:
Um, here is the definition of letter-spacing:normal

> The spacing is the normal spacing for the current font. Unlike a value of 0, this value allows the user agent to alter the space between characters in order to justify text.

PR #463, opened 2016-10-18
Update screenshot.png

melchoyce:
See #119


josephfusco:
@melchoyce Instead of

> Bringing your business site to life

should we do

> Bringing your business' site to life

? The first being literally a **business site**, while the latter is a **site** for your business.

laurelfulford:
@josephfusco Sorry, I think I merged just as you commented. Could you please create a new issue for your question, and we can iterate on the screenshot? Thank you!

Issue #450, opened 2016-10-18
iPad mini: Images on the front page badly pixelated


alaczek:
On iPad mini 2 retina in landscape mode both custom header image and panel images appear to be enlarged and badly pixelated:
![image](https://cloud.githubusercontent.com/assets/4550351/19464532/cc75e8aa-9548-11e6-9b37-1372d638361c.png)

![image](https://cloud.githubusercontent.com/assets/4550351/19464535/d0a07954-9548-11e6-9a68-4a35dc02ef34.png)

In portrait mode images display correctly:
![image](https://cloud.githubusercontent.com/assets/4550351/19464542/e407c6f0-9548-11e6-9488-37c329ff3478.png)

Original size of featured image above is 2400x1600.

Featured images on single posts/pages look good as well.

laurelfulford:
Looks like this is related to `background-attachment: fixed`. 

Altering the breakpoint for where this is 'turned off' should help.

karmatosed:
Moved over to trac: https://core.trac.wordpress.org/ticket/38395

PR #465, opened 2016-10-18
Turning off 'background-attachment: fixed' for iOS

laurelfulford:
Turning off the `background-attachment: fixed` style for iOS devices; it's not supported, and causes the images to be blown up to really huge, blurry sizes.

I steered away from sized-based `@media` queries, since iPad Pros are so large and it would cause the effect to be 'turned off' on laptop-sized screens, too. 

However, what I ended up with feels hacky to me - I'd be interested in suggestions others have for this. 

See #450.

### DIFF
--- a/style.css
+++ b/style.css
@@ -3751,16 +3751,21 @@ article.panel-placeholder {
 
 @media screen and ( min-width: 55em ) {
 
-	.custom-header-image {
-		background-attachment: fixed;
-	}
-
+	.custom-header-image,
 	.twentyseventeen-front-page:not(.no-header-image) .custom-header-image,
-	.home.blog:not(.no-header-image) .custom-header-image {
+	.home.blog:not(.no-header-image) .custom-header-image,
+	.panel-image {
 		background-attachment: fixed;
 	}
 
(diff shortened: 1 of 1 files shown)


josephfusco:
IMO I think we should go with a JS check rather than this wizardry.

josephfusco:
@laurelfulford I've been meaning to refactor the custom header image but haven't found time to yet.

The idea is using a `position: fixed` psuedo element without the fixed background attachment. This would also improve scrolling performance and in theory also fix #450.

laurelfulford:
> I think we should go with a JS check rather than this wizardry.

Thanks @josephfusco! I wasn't sure if that would be overkill, but it's probably easier to understand than this crazy CSS 🙂 

> The idea is using a position: fixed psuedo element without the fixed background attachment.

My only worry with this is that all the images on the front page use this, not just the custom header image. It sounds like this would work for the custom header space, but I'm not sure it's actually possible for anything further down the page.

davidakennedy:
@laurelfulford will take another stab at this after merge to Core.

PR #466, opened 2016-10-18
Update screenshot

melchoyce:
From “business site” to “business’ site”

Fixes #464

PR #468, opened 2016-10-18
Update screenshot 

melchoyce:
Messed up my PR. See #466.

This uses Photoshop's PNG-8 export instead of PNG-24.


josephfusco:
File size is a lot better here 👍  🎉  Looks great!

josephfusco (approved):

PR #470, opened 2016-10-18
Check to make sure entry-footer isn't empty

### LINKED ISSUE #314: Hide meta borders when meta is empty logged out

laurelfulford:
Adding a check to make sure `.entry-footer` will actually have stuff in it before writing it to the page, so we don't end up with an empty bordered space.

Fixes #314.

### DIFF
--- a/components/post/content-audio.php
+++ b/components/post/content-audio.php
@@ -86,9 +86,7 @@
 	</div><!-- .entry-content -->
 
 	<?php if ( is_single() ) : ?>
-		<footer class="entry-footer">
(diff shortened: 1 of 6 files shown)


melchoyce:
Looks good, but we'll need to add a border/spacing to pages that have comments:

screen shot 2016-10-18 at 9 23 21 pm

PR #473, opened 2016-10-18
Remove letter-spacing for all non-latin alphabets

melchoyce:
Fixing merge conflict in #460


joyously:
I think my comment might have been missed on the other PR (#460), in response to @josephfusco https://github.com/WordPress/twentyseventeen/pull/460#issuecomment-254638258
Here is the definition of letter-spacing:normal

>  The spacing is the normal spacing for the current font. Unlike a value of 0, this value allows the user agent to alter the space between characters in order to justify text.

Issue #453, opened 2016-10-18
iPhone 5s: Logo cut-off when using default custom header image


alaczek:
On iPhone 5s in portrait mode the logo is cut-off, if used together with custom header image:

![image](https://cloud.githubusercontent.com/assets/4550351/19464840/2e28b206-954b-11e6-8d16-dad563386f80.png)

Logo displays correctly when in landscape mode or if I remove the header image.

![image](https://cloud.githubusercontent.com/assets/4550351/19464830/1949747e-954b-11e6-8da1-14b0296f1a5e.png)

The size of the logo image used is 300x300px.

laurelfulford:
Was incidentally fixed in #481; closing!

PR #474, opened 2016-10-18
Updating header styles for custom header on all pages

laurelfulford:
Riffing on @ryelle's PR #414.

I just updated the header styles to make sure it doesn't end up being too short on subpages. Also simplified the header classes a bit - the theme was adding both `.has-header-image` and `.no-header-image` classes, and for the most part the latter was used with `:not(.no-header-image)` so I removed it. 

See #413.

### DIFF
--- a/assets/css/editor-style.css
+++ b/assets/css/editor-style.css
@@ -269,10 +269,14 @@ table {
 }
 
 a {
+	-webkit-box-shadow: inset 0 -1px 0 rgba(15, 15, 15, 1);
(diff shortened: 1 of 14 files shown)


melchoyce:
Looks good to me!

laurelfulford:
This should also fix #453.

davidakennedy:
@laurelfulford Can you refresh PR and update the default header text color?

laurelfulford:
Redid PR as #481.

PR #398, opened 2016-10-16
Add support for Arabic fonts 

melchoyce:
Add support for `ar`, `ary`, `azb`, `fa-IR`, `haz`, and `ps`. Adds to #59.

**Before**:

screen shot 2016-10-16 at 3 02 36 pm

**After**:

screen shot 2016-10-16 at 3 02 52 pm

PR #476, opened 2016-10-18
Add support for Urdu

grappler:
Adds support for Urdu

@ahmadawais Could you double check that this is right?

### DIFF
--- a/style.css
+++ b/style.css
@@ -602,7 +602,12 @@ html[lang="ps"] body,
 html[lang="ps"] button,
 html[lang="ps"] input,
 html[lang="ps"] select,
-html[lang="ps"] textarea {
+html[lang="ps"] textarea,
+html[lang="ur"] body,
+html[lang="ur"] button,
+html[lang="ur"] input,
+html[lang="ur"] select,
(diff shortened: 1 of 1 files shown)


ahmadawais:
@grappler I seldom work on translation these days, while `lang=ur` is def the right, I am not sure how `Tahoma` font will render Urdu. Noto on the other hand would be a better choice.

grappler:
@ahmadawais This is how it looks in a Arabic https://github.com/WordPress/twentyseventeen/pull/398 The problem with Noto we would need to load it from Google Fonts.

ahmadawais:
Sure go ahead then.  :)

melchoyce:
Thanks @grappler!

I don't see Urdu in the Site Language list:

screen shot 2016-10-18 at 9 17 31 pm
screen shot 2016-10-18 at 9 18 01 pm

Is there anything special I need to do to find it?

grappler:
Urdu is at 95% so that is why it is not showing up. https://translate.wordpress.org/locale/ur/default/wp/dev

melchoyce:
Thanks 👍

Issue #488, opened 2016-10-19
Incorrect theme tag - two-column


ernilambar:
Correct is `two-columns`.

PR #489, opened 2016-10-19
Fix incorrect theme tag

ernilambar:
##### What issue # is this related to?
#488
##### What is your WordPress.org username?

rabmalin
##### Description

Fix theme tag: `two-column` -> `two-columns`


style.css
```diff
@@ -4143,4 +4143,4 @@ article.panel-placeholder {
 	.site-footer {
 		padding: 0;
 	}
-}
\ No newline at end of file
+}
```
davidakennedy, line 4146: No need for the extra space here. Can you fix?

davidakennedy:
Fixed here: https://github.com/WordPress/twentyseventeen/commit/5d7ef4c5b7b8876b41a65b18cc84d75573fe4349

PR #491, opened 2016-10-19
Adds preconnect to Google Fonts.

### LINKED ISSUE #105: Implement preconnect to Google fonts

davidakennedy:
- See: https://core.trac.wordpress.org/ticket/38121
- A refresh of #106.

Fixes #105.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -150,7 +150,8 @@ function twentyseventeen_fonts_url() {
 	return esc_url_raw( $fonts_url );
 }
 
-/* Add preconnect for Google Fonts.
+/**
+ * Add preconnect for Google Fonts.
  *
  * @since Twenty Seventeen 1.0
  *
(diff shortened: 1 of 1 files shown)


davidakennedy:
Credit should go to `leobaiano` for this one. See #106.

swissspidy:
The PR looks good so far. According to the discussion at [#37171](https://core.trac.wordpress.org/ticket/37171) an array would need to passed though to add the `crossorigin` attribute. Example:

``` php
$urls[] = array(
    'href' => 'https://fonts.gstatic.com',
    'crossorigin',
);
```

This causes some notices on pre-4.7 installs, however, since this is a new feature. See https://core.trac.wordpress.org/changeset/38826.

leobaiano:
Tks.

PR #492, opened 2016-10-19
Remove Genericons

melchoyce:
Since we’re using Font Awesome now instead.


samikeijonen:
Note that we still have couple of icons from Genericons family. I would not remove that Copyright info from Readme file.

melchoyce:
Which icons? We should remove them.

samikeijonen:
I don't think we should remove them because they are used in a theme.

Atleast dropdown icon and arrows are from Genericon.

Why? Because there is no same kind of icons in font awesome that would fit
in the design.

I'm pretty sure I asked about it in early tickets:)

